PR #1, opened 2019-09-04
Added some files

akoushke:


.eslintrc.js
```diff
@@ -0,0 +1,51 @@
+module.exports = {
```
adamweeks: Could we use the same file format as the components and call this `.eslintrc`?

.eslintrc.js
```diff
@@ -0,0 +1,51 @@
...
+  parser: "babel-eslint",
+  extends: ["plugin:prettier/recommended"],
```
adamweeks: Let's use the webex eslint base config:
https://www.npmjs.com/package/@webex/eslint-config-base

akoushke: then we need to add that to `webex-components` too.

adamweeks: It already is. [eslint-config-react](https://github.com/webex/components/blob/master/.eslintrc#L4) includes the base config and adds react linting.

PR #3, opened 2019-09-19
Build - 1

akoushke:


.circleci/config.xml
```diff
@@ -0,0 +1,58 @@
...
+          paths: dist
+
+# execute the jobs in a orderly manner
+workflows:
+  setup_test_release:
+    jobs:
+      - install
+      - build:
+          requires:
+            - test
```
adamweeks: Requires test? I don't see that defined

akoushke: why is this not running it on the CI???

akoushke: okay addressed

.circleci/config.yml
```diff
@@ -0,0 +1,55 @@
...
+      - persist_to_workspace:
+          root: .
+          paths: dist
+
+# execute the jobs in a orderly manner
+workflows:
+  setup_test_release:
+    jobs:
+      - install
+      - build:
```
lalli-flores: Isn't this only supposed to run on `master`?

.circleci/config.yml
```diff
@@ -0,0 +1,55 @@
...
+          command: npm run build
+      - persist_to_workspace:
+          root: .
+          paths: dist
+
+# execute the jobs in a orderly manner
+workflows:
+  setup_test_release:
+    jobs:
+      - install
```
lalli-flores: Where's the test step?

.circleci/config.yml
akoushke: yeah, but I wanted to run it to test.

.circleci/config.yml
akoushke: will be added in JEST PR

PR #8, opened 2019-10-11
feat(WebexSdkAdapter): add base SDK adapter

akoushke:


src/WebexSDKAdapter.js
```diff
@@ -0,0 +1,13 @@
+export default class WebexSDKAdapter {
+  /**
+   * Creates a new instance of the WebexAdapter.
+   * The primary datasource model comes from the following structure:
+   * WebexSDKAdapter = {}
+   *
+   * @param {Object} datasource The primary datasource the json adapter will be using.
```
lalli-flores: I think we can call this `sdk` since we know the datasource for this adapter is the SDK.
Make sure to update the description accordingly

src/WebexSDKAdapter.js
```diff
@@ -0,0 +1,13 @@
+export default class WebexSDKAdapter {
+  /**
+   * Creates a new instance of the WebexAdapter.
```
lalli-flores:
```suggestion
   * Creates a new instance of the WebexSDKAdapter.
```

src/WebexSDKAdapter.js
```diff
@@ -0,0 +1,13 @@
...
+   * The primary datasource model comes from the following structure:
```
lalli-flores: `The data source for this adapter comes from an authenticated instance of the SDK.`

src/WebexSDKAdapter.js
```diff
@@ -0,0 +1,12 @@
+export default class WebexSDKAdapter {
+  /**
+   * Creates a new instance of the WebexSDKAdapter.
+   * The data source for this adapter comes from an authenticated instance of the SDK.
+   *
+   * @param {Object} sdk The primary sdk the json adapter will be using.
```
lalli-flores:
```suggestion
   * @param {Object} sdk The primary sdk the adapter will be using.
```

src/WebexSDKAdapter.js
```diff
@@ -0,0 +1,12 @@
+export default class WebexSDKAdapter {
+  /**
+   * Creates a new instance of the WebexSDKAdapter.
+   * The data source for this adapter comes from an authenticated instance of the SDK.
```
adamweeks: Should we add a link to the js-sdk project here?

lalli-flores:
> Do we need to update the existing tests?

I think we should just remove them right? There's really nothing to test on `WebexSDKAdapter`.

akoushke:
> > Do we need to update the existing tests?
> 
> I think we should just remove them right? There's really nothing to test on `WebexSDKAdapter`.

not having a test module will result into failure. This could be removed in Adam's PR

PR #9, opened 2019-10-17
Rooms adapter

adamweeks:
Builds on #8 

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-98330

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,8 @@
   },
   "devDependencies": {
     "@babel/core": "^7.6.0",
+    "@babel/plugin-transform-regenerator": "^7.4.5",
+    "@babel/plugin-transform-runtime": "^7.6.2",
     "@babel/preset-env": "^7.6.0",
     "@commitlint/cli": "^8.2.0",
     "@commitlint/config-conventional": "^8.2.0",
@@ -60,6 +62,10 @@
     "semantic-release": "^15.13.24"
   },
   "babel": {
+    "plugins": [
+      "@babel/plugin-transform-regenerator",
+      "@babel/plugin-transform-runtime"
+    ],
     "presets": [
       "@babel/preset-env"
     ]
--- a/package.json
+++ b/package.json
@@ -25,6 +25,8 @@
   "author": "devsupport@webex.com",
   "license": "MIT",
   "dependencies": {
+    "@webex/components": "^1.18.1",
+    "rxjs": "^6.5.3",
     "webex": "^1.79.0"
   },
   "devDependencies": {
@@ -47,6 +49,8 @@
     "jest": "^24.9.0",
     "jest-junit": "^8.0.0",
     "prettier": "^1.18.2",
+    "react": "^16.10.2",
+    "react-dom": "^16.10.2",
     "rollup": "^1.21.4",
     "rollup-plugin-babel": "^4.3.3",
     "rollup-plugin-commonjs": "^10.1.0",
--- a/src/RoomsSDKAdapter.js
+++ b/src/RoomsSDKAdapter.js
@@ -1,0 +1,62 @@
+import {Observable} from 'rxjs';
+import {RoomsAdapter} from '@webex/components';
+
+export const ROOM_UPDATED_EVENT = 'updated';
+
+/**
+ * The `RoomsSDKAdapter` is an implementation of the `RoomsAdapter` interface.
+ * This adapter utilizes the Webex JS SDK to fetch data about a room.
+ *
+ * @export
+ * @class RoomsSDKAdapter
+ * @extends {RoomsAdapter}
+ */
+export default class RoomsSDKAdapter extends RoomsAdapter {
+  /**
+   * Fetches the room data from the sdk and returns in the shape required by adapter.
+   *
+   * @param {string} ID
+   * @returns {Room}
+   * @memberof RoomsSDKAdapter
+   */
+  async fetchRoom(ID) {
+    const {id, title, type} = await this.datasource.rooms.get(ID);
+
+    return {
+      ID: id,
+      title,
+      type,
+    };
+  }
+
+  /**
+   * Returns an observable that emits room data of the given ID.
+   *
+   * @param {string} ID ID of room to get
+   * @returns {Observable.<Room>}
+   * @memberof RoomsSDKAdapter
+   */
+  getRoom(ID) {
+    return Observable.create((observer) => {
+      // Start listening for room changes
+      this.datasource.rooms.listen();
+      this.datasource.rooms.on(ROOM_UPDATED_EVENT, () => {
+        // Room has updates, fetch and send
+        this.fetchRoom(ID)
+          .then((room) => observer.next(room))
+          .catch((error) => observer.error(error));
+      });
+
+      // Get our initial room from the SDK
+      this.fetchRoom(ID)
+        .then((room) => observer.next(room))
+        .catch((error) => observer.error(error));
+
+      return () => {
+        // Cleanup when unsubscribing
+        this.datasource.rooms.stopListening();
+        this.datasource.rooms.off(ROOM_UPDATED_EVENT);
+      };
+    });
+  }
+}
--- a/src/RoomsSDKAdapter.test.js
+++ b/src/RoomsSDKAdapter.test.js
@@ -1,0 +1,77 @@
+import {isObservable} from 'rxjs';
+
+import RoomsSDKAdapter, {ROOM_UPDATED_EVENT} from './RoomsSDKAdapter';
+import createMockSDK from './__mocks__/sdk';
+
+describe('Rooms SDK Adapter', () => {
+  let roomsSDKAdapter;
+  let mockSDK, mockSDKRoom;
+
+  beforeEach(() => {
+    mockSDK = createMockSDK();
+    roomsSDKAdapter = new RoomsSDKAdapter(mockSDK);
+
+    mockSDKRoom = {
+      id: 'abc',
+      type: 'group',
+      title: 'mock room',
+    };
+  });
+
+  describe('getRoom() functionality', () => {
+    test('returns an observable', () => {
+      expect(isObservable(roomsSDKAdapter.getRoom())).toBeTruthy();
+    });
+
+    test('returns a room in a proper shape', (done) => {
+      mockSDK.rooms.get = jest.fn(() => Promise.resolve(mockSDKRoom));
+
+      roomsSDKAdapter.getRoom('id').subscribe((room) => {
+        expect(room).toEqual(
+          expect.objectContaining({
+            ID: mockSDKRoom.id,
+            type: mockSDKRoom.type,
+            title: mockSDKRoom.title,
+          })
+        );
+        done();
+      });
+    });
+
+    test('listens to room events when subscribing', (done) => {
+      mockSDK.rooms.get = jest.fn(() => Promise.resolve(mockSDKRoom));
+
+      roomsSDKAdapter.getRoom('id').subscribe(() => {
+        expect(mockSDK.rooms.on).toHaveBeenCalledWith(ROOM_UPDATED_EVENT, expect.any(Function));
+        done();
+      });
+    });
+
+    test('stops listening to events when unsubscribing', () => {
+      mockSDK.rooms.get = jest.fn(() => Promise.resolve(mockSDKRoom));
+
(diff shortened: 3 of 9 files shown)


src/RoomsSDKAdapter.js
```diff
@@ -0,0 +1,35 @@
...
+   * Returns an observable that emits room data of the given ID.
+   *
+   * @param {string} ID ID of room to get
+   * @returns {Observable.<Room>}
+   * @memberof RoomsSDKAdapter
+   */
+  getRoom(ID) {
+    return Observable.create(async (observer) => {
+      try {
+        const room = await this.sdk.rooms.get(ID);
```
lalli-flores: How would we handle room updates, such as updating the title of the room?

package.json
```diff
@@ -25,6 +25,11 @@
   "author": "devsupport@webex.com",
   "license": "MIT",
   "dependencies": {
+    "@webex/components": "^1.18.1",
+    "prop-types": "^15.7.2",
```
lalli-flores: I think `prop-types` is not necessary to add.

package.json
```diff
@@ -25,6 +25,11 @@
...
+    "react": "^16.10.2",
```
lalli-flores: I understand that this are peer dependencies of `@webex/components`, but we only really need it for testing, would it make more sense to add them as `devDependencies`?

src/RoomsSDKAdapter.js
adamweeks: The sdk does not provide live updates through the public api.

package.json
adamweeks: Sure

src/WebexSDKAdapter.js
```diff
@@ -0,0 +1,16 @@
...
+export default class WebexSDKAdapter {
+  /**
+   * Creates a new instance of the WebexSDKAdapter.
+   * The data source for this adapter comes from an authenticated instance of the SDK.
+   *
+   * @param {Object} sdk The primary sdk the adapter will be using.
+   * @todo implement and collect sub-adapters here in the constructor
+   */
+  constructor(sdk) {
+    this.sdk = sdk;
```
akoushke: don't think we would need this anymore. Had to add it to prevent displaying some eslint errors.

src/WebexSDKAdapter.js
```diff
@@ -0,0 +1,16 @@
+import RoomsSDKAdapter from './RoomsSDKAdapter';
+
+export default class WebexSDKAdapter {
+  /**
+   * Creates a new instance of the WebexSDKAdapter.
+   * The data source for this adapter comes from an authenticated instance of the SDK.
+   *
+   * @param {Object} sdk The primary sdk the adapter will be using.
+   * @todo implement and collect sub-adapters here in the constructor
```
akoushke: I think we should remove this comment in your PR

akoushke:
as i mentioned in the other PR, we should remove the `index.test.js` in this PR, since we have a legit test module now.

src/WebexSDKAdapter.js
adamweeks: I'm using `this.sdk` in the rooms adapter so we definitely need it.

package.json
adamweeks: img

It didn't work in node without it, but jest works fine with it. Removing.

package.json
```diff
@@ -58,7 +62,21 @@
...
       "@babel/preset-env"
-    ]
+    ],
+    "env": {
+      "test": {
+        "presets": [
+          [
+            "@babel/preset-env",
+            {
+              "targets": {
```
akoushke: any reason, we are setting the targets here?

src/RoomsSDKAdapter.test.js
```diff
@@ -0,0 +1,59 @@
+import {isObservable} from 'rxjs';
+
+import RoomsSDKAdapter from './RoomsSDKAdapter';
+
+describe('Rooms SDK Adapter Interface', () => {
+  let roomsSDKAdapter;
+  let mockSDK;
+
+  beforeEach(() => {
+    mockSDK = {
```
akoushke: shouldn't this be globally defined? we are going to need this mockSDK in basically every unit test module.

src/RoomsSDKAdapter.test.js
```diff
@@ -0,0 +1,59 @@
...
+      const mockSDKRoom = {
+        id: 'abc',
+        type: 'group',
+        title: 'mock room',
+      };
+
+      mockSDK.rooms.get = jest.fn(() => Promise.resolve(mockSDKRoom));
+
+      roomsSDKAdapter.getRoom('id').subscribe((room) => {
+        expect(room.ID).toBe(mockSDKRoom.id);
```
akoushke:
```js
expect(room).toEqual(mockSDKRoom)
```

perhaps?

src/RoomsSDKAdapter.test.js
```diff
@@ -0,0 +1,59 @@
...
+  });
+
+  describe('#getRoom()', () => {
+    test('returns an observable', () => {
+      expect(isObservable(roomsSDKAdapter.getRoom())).toBeTruthy();
+    });
+
+    test('returns a room in a proper shape', (done) => {
+      const mockSDKRoom = {
+        id: 'abc',
```
akoushke: ID

src/WebexSDKAdapter.test.js
```diff
@@ -0,0 +1,19 @@
...
+
+describe('Webex SDK Adapter Interface', () => {
+  let webexSDKAdapter;
+  const mockSDK = {};
+
+  beforeEach(() => {
+    webexSDKAdapter = new WebexSDKAdapter(mockSDK);
+  });
+
+  test('has a rooms adapter instance', () => {
```
akoushke: I think we can use this: https://jestjs.io/docs/en/expect#tobeinstanceofclass

package.json
adamweeks: Without setting a target of node, when running jest tests you get the following error:

>ReferenceError: regeneratorRuntime is not defined

This solution was provided in the following issue:
https://github.com/facebook/jest/issues/3126#issuecomment-483320742

src/RoomsSDKAdapter.test.js
adamweeks: You are right, we will need to mock the sdk in every test suite. We could have a helper function that mocks the sdk, or use your global definition. I'm chose to only mock the part of the sdk that is used by the RoomsAdapter so the test setup is very specific for the tests that are run.

src/RoomsSDKAdapter.test.js
adamweeks: The shape of the data that is returned from the SDK has `id` while the adapter expects `ID`. You can see in the implementation that we have to shape the data to match the adapter requirements so we can't do a direct comparison between the sdk object and the adapter object.

src/RoomsSDKAdapter.test.js
adamweeks: SDK returns `id`. Adapter returns `ID`.

package.json
akoushke: I'm tyring to figure out why we did not get that in webexComponent? maybe `pre-react` would resolve the issue?

adamweeks: I tried everything including installing "@babel/preset-react". Not sure if storybook does something "magical" for jest tests or not.

src/RoomsSDKAdapter.test.js
akoushke: I was just a bit hesitant about having multiple assertions in one test. However, I did some researches and learned that as long we follow this `Arrange, Act, Assert` pattern, it is fine to have multiple assertions. more infor: http://wiki.c2.com/?ArrangeActAssert

src/RoomsSDKAdapter.js
lalli-flores: How would we get live updates otherwise?

adamweeks: Nevermind, looks like JP added this feature to the rooms plugin a couple months ago. Will update the PR.

package.json
lalli-flores: If you read the [following comment](https://github.com/facebook/jest/issues/3126#issuecomment-519845717) after the fix you suggested, it says that it will fix the test runner but it won't properly transpile the bundle.

I ran into this issue before, and the solution seemed to be to add polyfills for babel (`core-js` and `runtime-generator`). See [@babel/polyfill](https://babeljs.io/docs/en/next/babel-polyfill.html) docs

src/RoomsSDKAdapter.js
```diff
@@ -0,0 +1,35 @@
+import {Observable} from 'rxjs';
+import {RoomsAdapter} from '@webex/components';
+
+export default class RoomsSDKAdapter extends RoomsAdapter {
+  constructor(sdk) {
+    super(sdk);
+
+    this.sdk = sdk;
```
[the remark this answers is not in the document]
adamweeks: Do you want me to use `this.datasource` or `this.sdk`? In the base adapter we are storing the sdk instance as `this.sdk` and for consistency with the base adapter I'm using it this way.

src/RoomsSDKAdapter.js
```diff
@@ -0,0 +1,35 @@
+import {Observable} from 'rxjs';
+import {RoomsAdapter} from '@webex/components';
+
+export default class RoomsSDKAdapter extends RoomsAdapter {
```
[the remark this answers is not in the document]
adamweeks: I'm working on one subtask, https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-98330 which has nothing to do with the activities.

src/RoomsSDKAdapter.test.js
```diff
@@ -0,0 +1,56 @@
...
+  let roomsSDKAdapter;
+  let mockSDK;
+
+  beforeEach(() => {
+    mockSDK = createMockSDK();
+
+    roomsSDKAdapter = new RoomsSDKAdapter(mockSDK);
+  });
+
+  describe('#getRoom()', () => {
```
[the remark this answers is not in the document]
adamweeks: It is a convention the sdk uses to indicate a function. If there is something else you'd prefer please comment

src/RoomsSDKAdapter.js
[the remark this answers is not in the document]
lalli-flores: I thought `this.sdk` was removed from the base adapter?

src/RoomsSDKAdapter.js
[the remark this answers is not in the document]
lalli-flores: Got it 👍🏻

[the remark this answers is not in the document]
adamweeks: If we call the `getRoomActivities` function on the adapter, it will 😄

akoushke:
could you please address these npm issues?
Screen Shot 2019-10-18 at 1 05 51 PM

src/RoomsSDKAdapter.js
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * Fetches the room data from the sdk and returns in the shape required by adapter.
+   *
+   * @param {string} ID
+   * @returns {Room}
+   * @memberof RoomsSDKAdapter
+   */
+  async fetchRoom(ID) {
+    const room = await this.datasource.rooms.get(ID);
```
akoushke:
```js
const {id, title, type} = await this.datasource.rooms.get(ID);
```

src/RoomsSDKAdapter.test.js
```diff
@@ -0,0 +1,85 @@
...
+      mockSDK.rooms.get = jest.fn(() => Promise.resolve(mockSDKRoom));
+
+      roomsSDKAdapter.getRoom('id').subscribe(() => {
+        expect(mockSDK.rooms.on).toHaveBeenCalledWith(ROOM_UPDATED_EVENT, expect.any(Function));
+        done();
+      });
+    });
+
+    test('stops listening to events when unsubscribing', () => {
+      const mockSDKRoom = {
```
akoushke: maybe have this defined in `beforeEach` phase?

package.json
```diff
@@ -56,6 +62,10 @@
     "semantic-release": "^15.13.24"
   },
   "babel": {
+    "plugins": [
```
akoushke: nice

src/RoomsSDKAdapter.js
adamweeks: I can fix this, but these kind of review comments should not be PR blockers.

PR #10, opened 2019-10-21
Rooms adapter integration tests

adamweeks:
Adding integration tests for rooms adapter, utilizing the Webex SDK.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,8 @@
     "release:dry-run": "semantic-release --dry-run",
     "start": "NODE_ENV=development nodemon scripts/start.js --exec babel-node",
     "test": "jest",
+    "test:integration": "jest \"^(?=.*integration.test.js).*\"",
+    "test:unit": "jest \"^(?=.*.test.js)(?!.*integration).*\"",
     "test:watch": "jest --watch",
     "test:coverage": "jest --coverage",
     "test:eslint": "eslint src/"
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "@semantic-release/changelog": "^3.0.4",
     "@semantic-release/git": "^7.0.16",
     "@webex/eslint-config-base": "^0.8.0",
+    "@webex/test-users": "^1.80.36",
     "babel-eslint": "^10.0.3",
     "conventional-changelog-angular": "^5.0.3",
     "dotenv": "^8.2.0",
@@ -68,7 +69,7 @@
     "webex": "^1.80.43"
   },
   "peerDependencies": {
-    "webex": "^1.79.0"
+    "webex": "^1.80.43"
   },
   "babel": {
     "plugins": [
--- a/src/RoomsSDKAdapter.integration.test.js
+++ b/src/RoomsSDKAdapter.integration.test.js
@@ -1,0 +1,83 @@
+import {createIntegrationTestUser, removeIntegrationTestUser} from './testHelper';
+import RoomsSDKAdapter from './RoomsSDKAdapter';
+
+describe('Rooms SDK Adapter', () => {
+  let roomsSDKAdapter, subscription, user;
+
+  // Since these are integration tests with live data,
+  // increase the async "idle" timeout so jest doesn't error early.
+  jest.setTimeout(30000);
+
+  beforeAll(async () => {
+    user = await createIntegrationTestUser();
+  });
+
+  beforeEach(() => {
+    roomsSDKAdapter = new RoomsSDKAdapter(user.sdk);
+  });
+
+  afterAll(async () => {
+    // This is done after unsubscribing, but since unsubscribe can't be async, manually call it.
+    await user.sdk.rooms.stopListening();
+
+    try {
+      await removeIntegrationTestUser(user);
+    } catch (reason) {
+      // eslint-disable-next-line no-console
+      console.warn('Failed to delete test user for Room SDK Adapter integration tests', reason);
+    }
+  });
+
+  describe('getRoom() returns', () => {
+    afterEach(() => {
+      subscription.unsubscribe();
+    });
+
+    test('a room in a proper shape', async (done) => {
+      const createdRoom = await user.sdk.rooms.create({title: 'Webex Teams Test Room'});
+
+      subscription = roomsSDKAdapter.getRoom(createdRoom.id).subscribe((room) => {
+        expect(room).toMatchObject({
+          ID: createdRoom.id,
+          title: createdRoom.title,
+        });
+        done();
+      });
+    });
+
+    test('an updated room title after subscribing', async (done) => {
+      const createdRoom = await user.sdk.rooms.create({title: 'Webex Teams Test Room'});
+      let hasUpdated = false;
+      const updatedTitle = 'Updated Test Title';
+
+      subscription = roomsSDKAdapter.getRoom(createdRoom.id).subscribe(async (room) => {
+        // The first subscription event will be the current state of the room.
+        if (!hasUpdated) {
+          // Update the room title so we get another subscription event.
+          await user.sdk.rooms.update({id: room.ID, title: updatedTitle});
+          hasUpdated = true;
+        } else {
+          // Once we've updated the title, the next event should include the updated title.
+          expect(room.title).toBe(updatedTitle);
+          done();
+        }
+      });
+    });
+
+    test('support for multiple subscriptions', async (done) => {
+      const createdRoom = await user.sdk.rooms.create({title: 'Webex Teams Test Room'});
+
+      subscription = roomsSDKAdapter.getRoom(createdRoom.id).subscribe();
+
+      const secondSubscription = roomsSDKAdapter.getRoom(createdRoom.id).subscribe((room) => {
+        expect(room).toMatchObject({
+          ID: createdRoom.id,
+          title: createdRoom.title,
+        });
+        done();
+      });
+
+      subscription.add(secondSubscription);
+    });
+  });
+});
--- a/src/testHelper.js
+++ b/src/testHelper.js
@@ -1,0 +1,76 @@
+import dotenv from 'dotenv';
+import {createTestUser, removeTestUser} from '@webex/test-users';
+import Webex from 'webex';
+
+// Used to get WEBEX_CLIENT_ID & WEBEX_CLIENT_SECRET for test user creation access
+dotenv.config();
+
+const URL_CONVO_SERVICE_INTEGRATION = 'https://conversation-intb.ciscospark.com/conversation/api/v1';
+const URL_ID_BROKER_INTEGRATION = 'https://idbrokerbts.webex.com';
+const URL_U2C_INTEGRATION = 'https://u2c-intb.ciscospark.com/u2c/api/v1';
+const SCOPE =
+  'Identity:OAuthClient webexsquare:get_conversation webexsquare:admin Identity:SCIM spark:people_read spark:rooms_read spark:rooms_write spark:memberships_read spark:memberships_write spark:messages_read spark:messages_write spark:applications_read spark:applications_write spark:teams_read spark:teams_write spark:team_memberships_read spark:team_memberships_write spark:bots_read spark:bots_write spark:kms';
+
+/**
+ * Creates an SDK instance for the given access token, pointed at the webex integration environment.
+ *
+ * @param {Object} token details about user token, generated from createTestUser
+ * @returns {Object} sdk instance
+ */
+export function createIntegrationSdkInstance(token) {
+  // Creating another instance of the token object because `new Webex` modifies the object passed to it.
+  const credentials = {...token};
+
+  return new Webex({
+    credentials,
+    config: {
+      // Set U2C to integration environment
+      services: {
+        u2c: URL_U2C_INTEGRATION,
+      },
+      logger: {
+        level: 'error',
+      },
+    },
+  });
+}
+
+/**
+ * Creates a test user in the integration webex environment.
+ *
+ * @returns {Object} user
+ * @returns {Object} user.sdk SDK Instance for the created user
+ */
+export async function createIntegrationTestUser() {
+  const user = await createTestUser({
+    idbrokerUrl: URL_ID_BROKER_INTEGRATION,
+    conversationServiceUrl: URL_CONVO_SERVICE_INTEGRATION,
+    scope: SCOPE,
+  });
+
+  // Pause for 5 seconds for CI
+  await new Promise((done) => setTimeout(done, 5000));
+
+  user.sdk = createIntegrationSdkInstance(user.token);
+
+  await user.sdk.internal.device.register();
+  await user.sdk.internal.mercury.connect();
+
+  return user;
+}
+
+/**
+ * Removes the test user from the integration environment.
+ *
+ * @param {Object} user created in #createIntegrationTestUser()
+ * @returns {Promise}
+ */
+export async function removeIntegrationTestUser(user) {
+  await user.sdk.internal.mercury.disconnect();
+  await user.sdk.internal.device.unregister();
+
+  return removeTestUser({
+    ...user,
+    conversationServiceUrl: URL_CONVO_SERVICE_INTEGRATION,
+  });
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -2726,6 +2726,63 @@
         "envify": "^4.1.0"
       }
     },
+    "@webex/test-users": {
+      "version": "1.80.36",
+      "resolved": "https://registry.npmjs.org/@webex/test-users/-/test-users-1.80.36.tgz",
+      "integrity": "sha512-Q232p+5PT6XvVzRMXxdrsi3RISSBVY7RVOmKCSP8MOsK8W5hZVKsE6MBB+QrWnTdULA/UFjDkSAk/Q36uvBSSA==",
+      "dev": true,
+      "requires": {
+        "@webex/http-core": "1.80.36",
+        "babel-runtime": "^6.26.0",
+        "btoa": "^1.2.1",
+        "envify": "^4.1.0",
+        "lodash": "^4.17.15",
+        "node-random-name": "^1.0.1",
+        "uuid": "^3.3.2"
+      },
+      "dependencies": {
+        "@webex/common": {
+          "version": "1.80.36",
+          "resolved": "https://registry.npmjs.org/@webex/common/-/common-1.80.36.tgz",
+          "integrity": "sha512-gq8X16aAMLSXaiHygLvV42Aj0r1MJBxSiBT8CXe3HqCasgXbJcBwJx4eTzw7kRZCw+AvAKC/oQPBDCS7bEbWIQ==",
+          "dev": true,
+          "requires": {
+            "@webex/common": "1.80.36",
+            "babel-runtime": "^6.26.0",
+            "backoff": "^2.5.0",
+            "core-decorators": "^0.20.0",
+            "envify": "^4.1.0",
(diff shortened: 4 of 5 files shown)


adamweeks:
@lalli-flores @InteractiveTimmy @akoushke ready for review

src/RoomsSDKAdapter.integration.test.js
```diff
@@ -0,0 +1,47 @@
...
+
+describe('Rooms SDK Adapter Interface', () => {
+  let roomsSDKAdapter;
+  let sdk;
+  let user;
+  let createdRoom;
+
+  beforeAll(async () => {
+    // Setting timeout higher for user creation and room creation
+    jest.setTimeout(15000);
```
lalli-flores: Since there is already a sleep in the user creation, is this also necessary? Why would creating an SDK instance would increase the idling?

src/RoomsSDKAdapter.integration.test.js
```diff
@@ -0,0 +1,47 @@
+import {removeTestUser} from '@webex/test-users';
+
+import {createIntegrationTestUser, createIntegrationSdkInstance} from './testHelper';
+import RoomsSDKAdapter from './RoomsSDKAdapter';
+
+describe('Rooms SDK Adapter Interface', () => {
```
```suggestion
describe('Rooms SDK Adapter', () => {
```

src/RoomsSDKAdapter.integration.test.js
```diff
@@ -0,0 +1,47 @@
...
+    jest.setTimeout(15000);
+
+    user = await createIntegrationTestUser();
+
+    sdk = createIntegrationSdkInstance(user.token);
+
+    createdRoom = await sdk.rooms.create({title: 'Webex Teams Test Room'});
+  });
+
+  afterAll(() =>
```
lalli-flores: Can we move this to the end of the block? e.g.

```
beforeAll()
beforeEach()

test

afterEach()
afterAll()

File: src/RoomsSDKAdapter.integration.test.js [hunk 4]
```diff
@@ -0,0 +1,47 @@
...
+      // eslint-disable-next-line no-console
+      console.warn('failed to delete test user', reason);
+    })
+  );
+
+  beforeEach(() => {
+    roomsSDKAdapter = new RoomsSDKAdapter(sdk);
+  });
+
+  describe('getRoom() functionality', () => {
```
lalli-flores: On the component repo, I setup the describe 
```suggestion
  describe('getRoom() returns', () => {
```
[See sample test file](https://github.com/webex/components/blob/master/src/components/WebexActivity/ActivityHeader.test.js#L24) in component repo

src/RoomsSDKAdapter.integration.test.js
```diff
@@ -0,0 +1,47 @@
...
+    test('returns a room in a proper shape', (done) => {
+      roomsSDKAdapter.getRoom(createdRoom.id).subscribe((room) => {
+        expect(room.ID).toBe(createdRoom.id);
```
lalli-flores: I thought we agreed to use [`toMatchObject`](https://jestjs.io/docs/en/expect#tomatchobjectobject) instead of multiple `expect`s

src/testHelper.js
```diff
@@ -0,0 +1,48 @@
+import dotenv from 'dotenv';
+import {createTestUser} from '@webex/test-users';
+import Webex from 'webex';
+
+// Used to get WEBEX_CLIENT_ID & WEBEX_CLIENT_SECRET for test user creation access
```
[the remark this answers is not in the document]
adamweeks: We didn't add one for the start script. It might be worthwhile to add a testing section to the README though.

package.json
```diff
@@ -13,6 +13,8 @@
     "release:dry-run": "semantic-release --dry-run",
     "start": "NODE_ENV=development nodemon scripts/start.js --exec babel-node",
     "test": "jest",
+    "test:integration": "jest \"^(?=.*integration.test.js).*\" --forceExit",
```
adamweeks: `--forceExit` was required due to the jest tests not exiting. Tried debugging with `--detectOpenHandles` with no luck.

package.json
```diff
@@ -13,6 +13,8 @@
...
+    "test:unit": "jest \"^(?=.*.test.js)(?!.*integration).*\"",
```
adamweeks: Thanks to @InteractiveTimmy for the regex magic

src/testHelper.js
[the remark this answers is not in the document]
lalli-flores: To be done in https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-104021

src/RoomsSDKAdapter.integration.test.js
lalli-flores: Ok, now I understand this is NOT the same thing as Javascript's `setTimeout` but I'm still confused as to what it does.
Does it set an _error_ timeout for test? e.g. "if the test doesn't terminate in X sec move on"
or does it set the interval to wait between tests? e.g. "test 1 executed ...wait 1.5 secs ... execute test 2, etc..."

src/RoomsSDKAdapter.integration.test.js
```diff
@@ -0,0 +1,46 @@
+import {removeTestUser} from '@webex/test-users';
+
+import {createIntegrationTestUser, createIntegrationSdkInstance} from './testHelper';
+import RoomsSDKAdapter from './RoomsSDKAdapter';
+
+describe('Rooms SDK Adapter', () => {
+  let roomsSDKAdapter;
```
lalli-flores: In other repos ([components](https://github.com/webex/components/blob/master/src/adapters/ActivitiesJSONAdapter.test.js#L5), [interfaces](https://github.com/webex/component-adapter-interfaces/blob/master/src/WebexAdapter.test.js#L4)) we have variables in a single line

package.json
lalli-flores: I don't like that we're using `--forceExit` because I feel like it's masking an actual issue

src/RoomsSDKAdapter.integration.test.js
```diff
@@ -0,0 +1,46 @@
...
+    createdRoom = await sdk.rooms.create({title: 'Webex Teams Test Room'});
+  });
+
+  beforeEach(() => {
+    roomsSDKAdapter = new RoomsSDKAdapter(sdk);
+  });
+
+  describe('getRoom() returns', () => {
+    test('a room in a proper shape', (done) => {
+      roomsSDKAdapter.getRoom(createdRoom.id).subscribe((room) => {
```
lalli-flores: I think the reason why the `jest` process doesn't finish is because we're not unsubscribing.

src/RoomsSDKAdapter.integration.test.js
```diff
@@ -0,0 +1,46 @@
...
+      });
+    });
+  });
+
+  afterEach(() => {
+    roomsSDKAdapter = null;
+  });
+
+  afterAll(() =>
+    removeTestUser(user).catch((reason) => {
```
lalli-flores: should we use `async/await` / `try/catch` syntax for consistency?
```suggestion
try {
  await removeTestUser(user)
} catch(reason) {
```

src/RoomsSDKAdapter.integration.test.js
```diff
@@ -0,0 +1,46 @@
...
+      // eslint-disable-next-line no-console
+      console.warn('failed to delete test user', reason);
```
lalli-flores: Can we have this be more specific so we know which test generated the error?
```suggestion
      console.warn('Failed to delete test user for Room SDK Adapter integration tests', reason);
```

package.json
lalli-flores: I think the issue might be we're not unsubscribing from `getRoom`, so the observable doesn't complete and the socket is still opened.

src/RoomsSDKAdapter.integration.test.js
lalli-flores: The only thing i found was [this jest plugin](https://www.npmjs.com/package/eslint-plugin-jest) but its rules say that they all should be on the top of the describe

src/testHelper.js
```diff
@@ -0,0 +1,48 @@
...
+export async function createIntegrationTestUser() {
+  const user = await createTestUser({
+    idbrokerUrl: 'https://idbrokerbts.webex.com',
+    conversationServiceUrl: 'https://conversation-intb.ciscospark.com/conversation/api/v1',
+    scope:
+      'Identity:OAuthClient webexsquare:get_conversation webexsquare:admin Identity:SCIM spark:people_read spark:rooms_read spark:rooms_write spark:memberships_read spark:memberships_write spark:messages_read spark:messages_write spark:applications_read spark:applications_write spark:teams_read spark:teams_write spark:team_memberships_read spark:team_memberships_write spark:bots_read spark:bots_write spark:kms',
+  });
+
+  // Pause for 5 seconds for CI
+  await new Promise((done) => setTimeout(done, 5000));
```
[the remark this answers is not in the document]
lalli-flores: I feel like this is part of why our tests are so flaky, any ideas on what we can do (obviously outside this PR) to have something more reliable?

src/RoomsSDKAdapter.integration.test.js
adamweeks: Jest will simply fail if an async operation doesn't complete in the 5 seconds that is defaulted. The purpose of this change is to bump up that wait time so the test user creation has time to complete.

Instead of the confusion that `jest.setTimeout` is causing, perhaps we should utilize the `timeout` second param for `beforeAll` as described here: https://jestjs.io/docs/en/api.html#beforeallfn-timeout

Thoughts @lalli-flores ?

src/RoomsSDKAdapter.integration.test.js
adamweeks: I actually prefer the grouping that way. All of the setup and teardown hooks are supposed to be before the tests. All of the jest examples (https://jestjs.io/docs/en/setup-teardown) show it this way. 

https://github.com/jest-community/eslint-plugin-jest/blob/HEAD/docs/rules/prefer-hooks-on-top.md

src/RoomsSDKAdapter.integration.test.js
adamweeks: I'll update the test to unsubscribe, but it still isn't finishing properly:

>Jest did not exit one second after the test run has completed.

>This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.

src/testHelper.js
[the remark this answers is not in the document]
adamweeks: Feel free to reach out to CI again, but we were told there was not a way to know when the created test user had been distributed to all services.

package.json
adamweeks: I've added the unsubscribe, but it still doesn't exit gracefully. The problem is definitely in the room subscription because it will exit fine after creating and destroying test users. 

@lalli-flores if this is truly a blocker for you, please help investigate.

src/RoomsSDKAdapter.integration.test.js
adamweeks: Since the room requests may take more than 5 seconds, I've abandoned the timeout param idea and moved the timeout to the top of the main describe.

src/RoomsSDKAdapter.integration.test.js
lalli-flores: Ok, we can open a different PR to add the jest linter

package.json
lalli-flores: I think we fixed this now? 🤞🏻

lalli-flores:
@adamweeks, can you rebase this so we can verify the changes in `RoomsSDKAdapter` solved the issue?

src/RoomsSDKAdapter.integration.test.js
```diff
@@ -0,0 +1,65 @@
...
+        done();
+      });
+    });
+
+    test('an updated room title after subscribing', async (done) => {
+      const createdRoom = await sdk.rooms.create({title: 'Webex Teams Test Room'});
+      let hasUpdated = false;
+      const updatedTitle = 'Updated Test Title';
+      const subscription = roomsSDKAdapter.getRoom(createdRoom.id).subscribe(async (room) => {
+        if (!hasUpdated) {
```
akoushke: new line here please.

src/RoomsSDKAdapter.integration.test.js
```diff
@@ -0,0 +1,65 @@
...
+        subscription.unsubscribe();
+        done();
+      });
+    });
+
+    test('an updated room title after subscribing', async (done) => {
+      const createdRoom = await sdk.rooms.create({title: 'Webex Teams Test Room'});
+      let hasUpdated = false;
+      const updatedTitle = 'Updated Test Title';
+      const subscription = roomsSDKAdapter.getRoom(createdRoom.id).subscribe(async (room) => {
```
akoushke: not too sure why we have this `isUpdated` check. if we subscribe to it first and then make changes, shouldn't that trigger the expect statement automatically?

src/testHelper.js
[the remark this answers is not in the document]
akoushke: would like to know what the alternatives are too. maybe we can start the conversation again. we are already at 20 seconds runtime. is there a specific person that we need to start this conversation with? @taymoork2

src/RoomsSDKAdapter.integration.test.js
```diff
@@ -0,0 +1,65 @@
+import {createIntegrationSdkInstance, createIntegrationTestUser, removeIntegrationTestUser} from './testHelper';
```
akoushke: it will be cool to have this setup and teardown logic outside so all of our integration test modules have access to internal stuff by only one device registration/ socket connection, etc. Unless, it's an implicit teardown logic. just like `stopListening` from the rooms module. that should be done for each test module at least for now.  But the explicti ones should live outside in  the higher scope (device registration, mercury connection).

src/testHelper.js
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Creates an SDK instance for the given access token, pointed at the webex integration environment.
+ *
+ * @export
+ * @param {Object} token details about user token, generated from createTestUser
+ * @returns {Object} sdk instance
+ */
+export function createIntegrationSdkInstance(token) {
+  // Creating another instance of the token object because `new Webex` modifies the object passed to it.
```
lalli-flores: Thank you for this comment! I was about to ask why did you make a copy of the token.
A little irked that the SDK does not treat inputs as immutable though 😐

src/testHelper.js
```diff
@@ -0,0 +1,70 @@
...
+const URL_CONVO_SERVICE_INTEGRATION = 'https://conversation-intb.ciscospark.com/conversation/api/v1';
+const URL_ID_BROKER_INTEGRATION = 'https://idbrokerbts.webex.com';
+const URL_U2C_INTEGRATION = 'https://u2c-intb.ciscospark.com/u2c/api/v1';
+const SCOPE =
+  'Identity:OAuthClient webexsquare:get_conversation webexsquare:admin Identity:SCIM spark:people_read spark:rooms_read spark:rooms_write spark:memberships_read spark:memberships_write spark:messages_read spark:messages_write spark:applications_read spark:applications_write spark:teams_read spark:teams_write spark:team_memberships_read spark:team_memberships_write spark:bots_read spark:bots_write spark:kms';
+
+/**
+ * Creates a test user in the integration webex environment.
+ *
+ * @export
```
lalli-flores: Is `@export` a JS Doc tag? I couldn't find it.
There is _@exports_, but JS Docs's [example section](https://jsdoc.app/tags-exports.html) says it's totally unnecessary

src/RoomsSDKAdapter.integration.test.js
```diff
@@ -0,0 +1,65 @@
...
+        } else {
+          expect(room.title).toBe(updatedTitle);
+          subscription.unsubscribe();
+          done();
+        }
+      });
+    });
+  });
+
+  afterAll(async () => {
```
lalli-flores: Should we have tests for multiple subscriptions?

src/RoomsSDKAdapter.integration.test.js
lalli-flores: I think this goes back to our conversation of having `connect/disconnect` at the `WebexSDKAdapter` level

package.json
```diff
@@ -44,6 +46,7 @@
     "@semantic-release/changelog": "^3.0.4",
     "@semantic-release/git": "^7.0.16",
     "@webex/eslint-config-base": "^0.8.0",
+    "@webex/test-users": "^1.80.36",
```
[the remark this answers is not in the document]
adamweeks: Yes, it is not included in the `webex` package.

src/RoomsSDKAdapter.integration.test.js
adamweeks: This is valid code and has no eslint violations.

src/RoomsSDKAdapter.integration.test.js
adamweeks: I'll add some comments, but this is due to the first `next` being the current state of the room and then we get another subscription update after the room title changes. Not sure how else you want me to do it.

src/testHelper.js
adamweeks: My jsdoc generator plugin adds them by default. I will remove if they bother you.
https://marketplace.visualstudio.com/items?itemName=joelday.docthis

src/testHelper.js
adamweeks: Agreed, that was a fun bug to chase down. If you used the same token to try to delete the test user, fields were missing.

adamweeks: Created a tech debt story for the SDK: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-105730

src/testHelper.js
lalli-flores: I see, weird that they add a tag that doesn't exist.
Let's keep our doc strings minimal, so i'd appreciate if you can remove them 👍

adamweeks: They have already been removed 😄

src/testHelper.js
```diff
@@ -0,0 +1,76 @@
...
+  const user = await createTestUser({
+    idbrokerUrl: URL_ID_BROKER_INTEGRATION,
+    conversationServiceUrl: URL_CONVO_SERVICE_INTEGRATION,
+    scope: SCOPE,
+  });
+
+  // Pause for 5 seconds for CI
+  await new Promise((done) => setTimeout(done, 5000));
+
+  user.sdk = createIntegrationSdkInstance(user.token);
```
lalli-flores: Interesting choice, adding the SDK instance to the user object.
I think this works, hope it doesn't cause confusion in the future.

src/RoomsSDKAdapter.integration.test.js
```diff
@@ -0,0 +1,88 @@
...
+          done();
+        }
+      });
+    });
+
+    describe('support for multiple subscriptions', () => {
+      let secondSubscription;
+
+      afterEach(() => {
+        secondSubscription.unsubscribe();
```
lalli-flores: Creative workaround to unsubscribing!
However, a cleaner solution would be to "add" new subscription to an already existing one (e.g. `subscription`). Like so:
```
subscription.add(secondSubscription);
```
The first unsubscribe will take care of doing un-subscription of all added subscriptions. See [`Subscription.add`](https://rxjs-dev.firebaseapp.com/api/index/class/Subscription#add-)

src/testHelper.js
adamweeks: I stole the idea from the SDK tests!

PR #11, opened 2019-10-21
fix(rollup): add babel and react configs

adamweeks:
runtimeHelpers were added to the base babel config in order to get the tests working with observables. 

Also added react rollup configs due to base adapters being part of the components repo.


akoushke:
we are getting `exceeded bundle size` from our very own `webexComponets` library. not a blocker though.

adamweeks:
> FYI, we are getting `exceeded bundle size` from our very own `webexComponets` library. not a blocker though.

@akoushke yea, I saw that. The bundle size might actually be from the `webex` import since it is quite large.

lalli-flores:
Did we verify bundle works?

PR #13, opened 2019-10-24
Export Default Webex Adapter

akoushke:
To trigger the release process.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,1 +1,3 @@
-export {default as WebexSDKAdapter} from './WebexSDKAdapter';
+import WebexSDKAdapter from './WebexSDKAdapter';
+
+export default WebexSDKAdapter;
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ import '@webex/components/dist/webexComponents.css';
 
 import React from 'react';
 import ReactDOM from 'react-dom';
-import {WebexSDKAdapter} from '@webex/sdk-component-adapter';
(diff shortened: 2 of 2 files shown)


src/index.js
```diff
@@ -1 +1,2 @@
 export {default as WebexSDKAdapter} from './WebexSDKAdapter';
+export {default as RoomsSDKAdapter} from './RoomsSDKAdapter';
```
adamweeks: Should `WebexSDKAdapter` be the default export?

That way one could:
```js
import WebexSDKAdapter from '@webex/sdk-component-adapter';
```

Or if someone was building their own adapter but wanted to use the RoomsSDKAdapter they could:

```js
import {RoomsSDKAdapter} from '@webex/sdk-component-adapter';

const myAdapter = {roomsAdapter: RoomsSDKAdapter};
```

lalli-flores: I like the idea of making `WebexSDKAdapter` a default export, but I still don’t think is a good idea to expose the sub-adapter.
My worry is that developers may use an “incomplete adapter” with components that might need more data that than of what is in the custom adapter wrapper.
This is why we use the “wrapper”. What sub-adapters a component uses is transparent, meaning, is not obvious/shown to users, so I feel not having a complete adapter will cause more harm than good.

adamweeks: Sounds good. We can revisit if we ever actually get that type of feature request. 

@akoushke can you update this to be the default export?

PR #14, opened 2019-10-25
Setup Dev Environment

akoushke:
ATM, `eslint` does not lint the scripts and I don't mind that. I'm assuming this `start.js` file will never be committed again.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
   "author": "devsupport@webex.com",
   "license": "MIT",
   "dependencies": {
-    "@webex/components": "^1.18.1",
+    "@webex/component-adapter-interfaces": "^1.2.1",
     "rxjs": "^6.5.3"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "release": "semantic-release",
     "release:debug": "semantic-release --debug",
     "release:dry-run": "semantic-release --dry-run",
+    "start": "NODE_ENV=development nodemon scripts/start.js --exec babel-node",
     "test": "jest",
     "test:watch": "jest --watch",
     "test:coverage": "jest --coverage",
@@ -26,14 +27,15 @@
   "license": "MIT",
   "dependencies": {
     "@webex/components": "^1.18.1",
-    "rxjs": "^6.5.3",
-    "webex": "^1.79.0"
+    "rxjs": "^6.5.3"
   },
   "devDependencies": {
-    "@babel/core": "^7.6.0",
+    "@babel/cli": "^7.6.4",
+    "@babel/core": "^7.6.4",
+    "@babel/node": "^7.6.3",
     "@babel/plugin-transform-regenerator": "^7.4.5",
     "@babel/plugin-transform-runtime": "^7.6.2",
-    "@babel/preset-env": "^7.6.0",
+    "@babel/preset-env": "^7.6.3",
     "@commitlint/cli": "^8.2.0",
     "@commitlint/config-conventional": "^8.2.0",
     "@semantic-release/changelog": "^3.0.4",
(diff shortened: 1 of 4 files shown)


scripts/start.js
```diff
@@ -0,0 +1,19 @@
...
+if (process.env.WEBEX_ACCESS_TOKEN) {
+  const webex = new Webex({
+    credentials: process.env.WEBEX_ACCESS_TOKEN,
+  });
+
+  // eslint-disable-next-line no-unused-vars
+  const webexSDKAdapter = new WebexSDKAdapter(webex);
+} else {
+  // eslint-disable-next-line no-console
+  console.error('Please set a valid WEBEX_ACCESS_TOKEN environment in `env.development` file at the root level.');
```
adamweeks: Can we `return process.exit(1);` here so the node script knows it failed?

akoushke: then I could get rid of the `else` condition. just have a quick `if` statement at top.

scripts/start.js
```diff
@@ -0,0 +1,19 @@
+import * as dotenv from 'dotenv';
+import Webex from 'webex';
+
+import WebexSDKAdapter from '../src/WebexSDKAdapter';
+
+// Allows convict module to read the corresponding configuration
+dotenv.config({path: `${__dirname}/../.env.development`});
```
adamweeks: Can we just have the file loaded from `.env` and then all we have to do is `dotenv.config()`

scripts/start.js
```diff
@@ -0,0 +1,19 @@
+import * as dotenv from 'dotenv';
+import Webex from 'webex';
+
+import WebexSDKAdapter from '../src/WebexSDKAdapter';
+
+// Allows convict module to read the corresponding configuration
```
adamweeks: Not sure what "convict" means.

scripts/start.js
akoushke: I liked having a dev environment as a separate environment to not mess things up later on, in case we need actual env variables.

akoushke: I can change it to what you want, if you still think it's not necessary.

package.json
```diff
@@ -25,15 +26,18 @@
   "author": "devsupport@webex.com",
   "license": "MIT",
   "dependencies": {
-    "@webex/components": "^1.18.1",
+    "@webex/component-adapter-interfaces": "^1.2.1",
```
lalli-flores: Can you have the install/remove from components repot to interface repo on its own commit?

package.json
```diff
@@ -25,15 +26,18 @@
...
+    "dotenv": "^8.2.0",
```
lalli-flores: This is a dev dependency

package.json
```diff
@@ -25,15 +26,18 @@
...
     "rxjs": "^6.5.3",
     "webex": "^1.79.0"
```
lalli-flores: We should make `webex` a peer dependency

scripts/start.js
```diff
@@ -0,0 +1,21 @@
...
+
+  process.exit(1);
+}
+
+const webex = new Webex({
+  credentials: process.env.WEBEX_ACCESS_TOKEN,
+});
+
+// eslint-disable-next-line no-unused-vars
+const webexSDKAdapter = new WebexSDKAdapter(webex);
```
lalli-flores: Can you add a comment after this? something like

```
--------------------------------------------------------------------------------
// Add your test/development code here

package.json
akoushke: this can not be a peer dependency in this repo. we are utilizing it in our scripts now.

akoushke: I will change it a `devDep`

lalli-flores: I'm ok with that, we don't want to bundle our implementation with the SDK.

scripts/start.js
lalli-flores: It's a [tool](https://www.npmjs.com/package/convict) we used in previous projects, prob copy-paste issue

PR #18, opened 2019-11-06
feat(PeopleSdkAdapter): implement adapter

akoushke:
People SDK Adapter

### DIFF
--- a/src/PeopleSDKAdapter.js
+++ b/src/PeopleSDKAdapter.js
@@ -1,0 +1,106 @@
+import {concat, from, fromEvent} from 'rxjs';
+import {flatMap, filter, finalize, map, publishReplay, refCount} from 'rxjs/operators';
+import {deconstructHydraId} from '@webex/common';
+import {PeopleAdapter, PersonStatus} from '@webex/component-adapter-interfaces';
+
+const USER_PRESENCE_UPDATE_EVENT = 'event:apheleia.subscription_update';
+
+/**
+ * The `PeopleSDKAdapter` is an implementation of the `PeopleAdapter` interface.
+ * This adapter utilizes the Webex JS SDK to fetch data about a person.
+ *
+ * @class PeopleSDKAdapter
+ * @extends {PeopleAdapter}
+ */
+export default class PeopleSDKAdapter extends PeopleAdapter {
+  constructor(datasource) {
+    super(datasource);
+
+    this.getPersonObservables = {};
+  }
+
+  /**
+   * Returns a proper Enum key for a given PersonStatus value
+   *
+   * @private
+   * @param {string} status  Person status from Apheleia service.
+   * @returns {string} PersonStatus
+   * @memberof PeopleSDKAdapter
+   */
+  getStatus(status) {
+    return Object.keys(PersonStatus).find((key) => PersonStatus[key] === status);
+  }
+
+  /**
+   * Fetches the person data from the sdk and returns in the shape required by adapter.
+   *
+   * @private
+   * @param {string} ID  ID of the person for which to fetch data
+   * @returns {Person}
+   * @memberof PeopleSDKAdapter
+   */
+  async fetchPerson(ID) {
+    const {id, emails, displayName, firstName, lastName, avatar, orgId} = await this.datasource.people.get(ID);
+
+    return {
+      ID: id,
+      emails,
+      displayName,
+      firstName,
+      lastName,
+      avatar,
+      orgID: orgId,
+    };
+  }
+
+  /**
+   * Returns an observable that emits person data of the given ID.
+   *
+   * @public
+   * @param {string} ID ID of person to get.
+   * @returns {Observable.<Person>}
+   * @memberof PeopleSDKAdapter
+   */
+  getPerson(ID) {
+    if (!(ID in this.getPersonObservables)) {
+      const personUUID = deconstructHydraId(ID).id;
+      const person$ = from(this.fetchPerson(ID));
+
+      // Subscribe to `Apheleia` internal service to listen for status changes
+      // And update the person object with status response from the subscription
+      const personWithStatus$ = from(this.datasource.internal.presence.subscribe(personUUID)).pipe(
+        map((data) => data.responses[0].status.status),
+        flatMap((status) => person$.pipe(map((person) => ({...person, status: this.getStatus(status)}))))
+      );
+
+      // Listen for status changes for the given person ID after subscription to service
+      const statusUpdate$ = fromEvent(this.datasource.internal.mercury, USER_PRESENCE_UPDATE_EVENT).pipe(
+        filter((event) => event.data.subject === personUUID),
+        map((event) => this.getStatus(event.data.status))
+      );
+
+      // Update the person status after each change emitted from the event
+      const personUpdate$ = person$.pipe(
+        flatMap((person) => statusUpdate$.pipe(map((status) => ({...person, status}))))
+      );
+
+      // Fetch the original person data on the first run
+      // and updated object after each status change
+      const getPerson$ = concat(personWithStatus$, personUpdate$).pipe(
+        finalize(() => {
+          // Unsubscribe from `Apheleia` internal service
+          // when there are no more subscriptions.
+          this.datasource.internal.presence.unsubscribe(personUUID);
+          delete this.getPersonObservables[ID];
+        })
+      );
+
+      this.getPersonObservables[ID] = getPerson$.pipe(
+        publishReplay(1),
+        refCount()
+      );
+    }
+
+    return this.getPersonObservables[ID];
+  }
+}
--- a/src/PeopleSDKAdapter.test.js
+++ b/src/PeopleSDKAdapter.test.js
@@ -1,0 +1,79 @@
+import {isObservable} from 'rxjs';
+
+import PeopleSDKAdapter from './PeopleSDKAdapter';
+import createMockSDK from './__mocks__/sdk';
+
+describe('People SDK Adapter', () => {
+  let peopleSDKAdapter, mockSDK, personID;
+
+  beforeEach(() => {
+    mockSDK = createMockSDK();
+    peopleSDKAdapter = new PeopleSDKAdapter(mockSDK);
+    personID = 'personID';
+  });
+
+  describe('getPerson()', () => {
+    test('returns an observable', () => {
+      expect(isObservable(peopleSDKAdapter.getPerson(personID))).toBeTruthy();
+    });
+
+    test('returns a person in a proper shape', (done) => {
+      peopleSDKAdapter.getPerson(personID).subscribe((person) => {
+        expect(person).toEqual({
+          ID: 'id',
+          emails: ['email@cisco.com'],
+          displayName: 'Webex Components',
+          firstName: 'Webex',
+          lastName: 'Components',
+          avatar: 'avatar',
+          orgID: 'orgID',
+          status: 'ACTIVE',
+        });
+        done();
+      });
+    });
+
+    test('stops listening to events when unsubscribing', () => {
+      const subscription = peopleSDKAdapter.getPerson(personID).subscribe();
+
+      subscription.unsubscribe();
+      expect(mockSDK.internal.presence.unsubscribe).toHaveBeenCalled();
+    });
+
+    test('throws error on failed person fetch request', (done) => {
+      const errorMessage = 'a proper people error message';
+
+      personID = 'invalid personID';
+      peopleSDKAdapter.fetchPerson = jest.fn(() => Promise.reject(new Error(errorMessage)));
+
+      peopleSDKAdapter.getPerson(personID).subscribe(
+        () => {},
+        (error) => {
+          expect(error.message).toBe(errorMessage);
+          done();
+        }
+      );
+    });
+
+    test('throws error on failed presence update subscription', (done) => {
+      const errorMessage = 'a proper subscription error message';
+
+      personID = 'invalid personID';
+      mockSDK.internal.presence.subscribe = jest.fn(() => Promise.reject(new Error(errorMessage)));
+
+      peopleSDKAdapter.getPerson(personID).subscribe(
+        () => {},
+        (error) => {
+          expect(error.message).toBe(errorMessage);
+          done();
+        }
+      );
+    });
+  });
+
+  afterEach(() => {
+    mockSDK = null;
+    peopleSDKAdapter = null;
+    personID = null;
+  });
+});
--- a/README.md
+++ b/README.md
@@ -45,10 +45,15 @@ import '@webex/components/dist/webexComponents.css';
 
 import React from 'react';
 import ReactDOM from 'react-dom';
+import Webex from 'webex';
 import WebexSDKAdapter from '@webex/sdk-component-adapter';
 import {WebexAvatar} from '@webex/components';
 
-const adapter = new WebexSDKAdapter();
+const webex = new Webex({
+  credentials: <YOUR WEBEX ACCESS_TOKEN>,
+});
+
+const adapter = new WebexSDKAdapter(webex);
 
 ReactDOM.render(
   <WebexDataProvider adapter={adapter} >
--- a/package-lock.json
+++ b/package-lock.json
@@ -2181,12 +2181,12 @@
       "dev": true
     },
     "@webex/common": {
-      "version": "1.80.36",
-      "resolved": "https://registry.npmjs.org/@webex/common/-/common-1.80.36.tgz",
-      "integrity": "sha512-gq8X16aAMLSXaiHygLvV42Aj0r1MJBxSiBT8CXe3HqCasgXbJcBwJx4eTzw7kRZCw+AvAKC/oQPBDCS7bEbWIQ==",
+      "version": "1.80.43",
+      "resolved": "https://registry.npmjs.org/@webex/common/-/common-1.80.43.tgz",
+      "integrity": "sha512-1SqmonPlW5mvbRmqb6C4GVfNxKK47TBabzyoIaLPxDE2ZQ8gOimcyG2/OfyNtwTx4hDG7FifDm3lRQk48lfCbg==",
       "dev": true,
       "requires": {
-        "@webex/common": "1.80.36",
+        "@webex/common": "1.80.43",
         "babel-runtime": "^6.26.0",
         "backoff": "^2.5.0",
         "core-decorators": "^0.20.0",
@@ -2196,9 +2196,9 @@
       }
     },
     "@webex/common-timers": {
-      "version": "1.80.36",
-      "resolved": "https://registry.npmjs.org/@webex/common-timers/-/common-timers-1.80.36.tgz",
-      "integrity": "sha512-1nUj7wesUkUfoq5fYHaAPAGT4IwN3b5SnuYCYYyU6ngcSrbArid9o5nlTXcoPBsMEVUjZUrnsIpPEP5EcmNv/A==",
+      "version": "1.80.43",
+      "resolved": "https://registry.npmjs.org/@webex/common-timers/-/common-timers-1.80.43.tgz",
+      "integrity": "sha512-XrPePfK36fqAoARX1M34TXHKaKR9/x3+b7tV0j3WT+/GmpAb40dli7HAncezVgTatc1RoWbVRXK2qIkONWuRJQ==",
       "dev": true,
       "requires": {
         "envify": "^4.1.0"
@@ -2216,9 +2216,9 @@
       "dev": true
     },
     "@webex/helper-html": {
-      "version": "1.80.36",
-      "resolved": "https://registry.npmjs.org/@webex/helper-html/-/helper-html-1.80.36.tgz",
-      "integrity": "sha512-kvn8H1DsbQoP9qulu/Jr5s0svL6XZY2DAV4W/K3pH98mbhhI71F/nYDMXIjKmMilR16gkaQOVcrgt3HFFftBnQ==",
+      "version": "1.80.43",
+      "resolved": "https://registry.npmjs.org/@webex/helper-html/-/helper-html-1.80.43.tgz",
+      "integrity": "sha512-TFtDSa/Wmjj5YgPwfxHJtYLI9lMShH1XXJ/Ja4Cf/fzWuQb5/hh/GdNJoJv7X6bwvplOmNZaaExjf8R9+Um5gA==",
       "dev": true,
       "requires": {
         "babel-runtime": "^6.26.0",
@@ -2227,12 +2227,12 @@
       }
     },
     "@webex/helper-image": {
-      "version": "1.80.36",
-      "resolved": "https://registry.npmjs.org/@webex/helper-image/-/helper-image-1.80.36.tgz",
-      "integrity": "sha512-Yi6SoTW/LMKxFdR2xJr6ZebrS/9uQGaIcTT+ZHw1IKwbN/H+Ia5vFRjHb5cniI9fxRctLdKzC0YTbWzQWT99cw==",
+      "version": "1.80.43",
+      "resolved": "https://registry.npmjs.org/@webex/helper-image/-/helper-image-1.80.43.tgz",
+      "integrity": "sha512-7fRu5mP6IwPIPAJBsPqoVZTpkTv9j9Dnx6W00e3Y/NgFe3c5+haZVBSwPuZx90wyH4aJrffwRsFCE1HyEoaHgQ==",
       "dev": true,
       "requires": {
-        "@webex/http-core": "1.80.36",
+        "@webex/http-core": "1.80.43",
         "babel-runtime": "^6.26.0",
         "envify": "^4.1.0",
         "exif": "^0.6.0",
@@ -2242,12 +2242,12 @@
       }
     },
     "@webex/http-core": {
-      "version": "1.80.36",
-      "resolved": "https://registry.npmjs.org/@webex/http-core/-/http-core-1.80.36.tgz",
-      "integrity": "sha512-yUvtouB1y/5NzMqiAUu6+izdavReW6OnSX4SUJ+tOqBo0kNVOqkWeQf+OKgQfchXxYS2PFZozKW8tuc451UGUw==",
+      "version": "1.80.43",
+      "resolved": "https://registry.npmjs.org/@webex/http-core/-/http-core-1.80.43.tgz",
+      "integrity": "sha512-x+fP/IaT8nvbhhYeinzI51kVkHCzcrBh1Jpvdr1liM6C702FWbl8GOYRMB7RqZpXHrT1EB+g1coe6kxN/IWZtw==",
       "dev": true,
       "requires": {
-        "@webex/common": "1.80.36",
+        "@webex/common": "1.80.43",
         "babel-runtime": "^6.26.0",
         "envify": "^4.1.0",
         "file-type": "^3.9.0",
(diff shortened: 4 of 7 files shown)


adamweeks:
Let's standardize on package updates commit messages being:
`fix(package): update webex dependency` instead of `build(webex): update dependency`

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,107 @@
...
+ * @export
+ * @class PeopleSDKAdapter
+ * @extends {PeopleAdapter}
+ */
+export default class PeopleSDKAdapter extends PeopleAdapter {
+  /**
+   * Returns a proper Enum key for a given PersonStatus value
+   *
+   * @param {string} status
+   * @returns {PersonStatus.<String>}
```
adamweeks: I'm not sure I understand what the type is supposed to be. Shouldn't this just be:
```suggestion
   * @returns {String} PersonStatus
```
since `PersonStatus` isn't an actual type

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,107 @@
...
+   */
+  getPersonUUID(ID) {
+    return Buffer.from(ID, 'base64')
+      .toString()
+      .split('/')[4];
+  }
+
+  async getPersonPresence(ID) {
+    try {
+      const {statusList} = await this.datasource.internal.presence.list([this.getPersonUUID(ID)]);
```
adamweeks: You can use `presence.get` since we are only getting one person.

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,107 @@
...
+    // eslint-disable-next-line consistent-return
+    return Observable.create((observer) => {
+      if (!ID) {
+        observer.error(new Error(`Could not find person with ID "${ID}"`));
+      } else {
+        this.fetchPerson(ID)
+          .then((person) => {
+            observer.next(person);
+
+            this.datasource.internal.mercury.on(USER_PRESENCE_UPDATE_EVENT, (event) => {
```
adamweeks: You haven't done a `presence.subscribe` for this person, so you will not get any of these events sent to your device.

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,107 @@
...
+        emails,
+        displayName,
+        firstName,
+        lastName,
+        avatar,
+        orgID: orgId,
+        status: await this.getPersonPresence(ID),
+      };
+    } catch (error) {
+      throw new Error(error);
```
adamweeks: Catching and throwing the same error seems redundant. If we're not doing any error handling, we don't really need the try/catch block.

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,107 @@
...
+            this.datasource.internal.mercury.on(USER_PRESENCE_UPDATE_EVENT, (event) => {
+              observer.next({...person, status: this.getStatus(event.data.status)});
+            });
+          })
+          .catch((error) => {
+            observer.error(error);
+          });
+
+        return () => {
+          // Cleanup when subscription
```
adamweeks: You'll want to `presence.unsubscribe` here as well.

src/PeopleSDKAdapter.test.js
```diff
@@ -0,0 +1,115 @@
...
+    test('returns an observable', () => {
+      expect(isObservable(peopleSDKAdapter.getPerson())).toBeTruthy();
+    });
+
+    test('returns a person in a proper shape', (done) => {
+      mockSDK.people.get = jest.fn(() => Promise.resolve(mockSDKPerson));
+      mockSDK.internal.presence.list = jest.fn(() => Promise.resolve({statusList: [{status: 'active'}]}));
+
+      peopleSDKAdapter.getPerson('id').subscribe((person) => {
+        expect(person).toEqual(
```
adamweeks: @lalli-flores discovered that [toMatchObject](https://jestjs.io/docs/en/expect#tomatchobjectobject) cleans up the expectation block a bit. We should standardize on that.

src/PeopleSDKAdapter.js
akoushke: I'm not following this.

src/PeopleSDKAdapter.js
akoushke: why are we not handling errors if we can? I'm not following.

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,107 @@
...
+      return this.getStatus(statusList[0].status);
+    } catch (error) {
+      throw new Error(error);
+    }
+  }
+
+  /**
+   * Fetches the person data from the sdk and returns in the shape required by adapter.
+   *
+   * @param {string} ID
```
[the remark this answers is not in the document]
akoushke: this is not the correct shape that RoomsAdapter follows. I just followed what we had in the RoomsAdapter.

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,107 @@
...
+   * @memberof PeopleSDKAdapter
+   */
+  getPerson(ID) {
+    // eslint-disable-next-line consistent-return
+    return Observable.create((observer) => {
+      if (!ID) {
+        observer.error(new Error(`Could not find person with ID "${ID}"`));
+      } else {
+        this.fetchPerson(ID)
+          .then((person) => {
```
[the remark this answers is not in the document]
akoushke: still working on making this work with `async/await`

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,119 @@
...
+   * @param {string} status
+   * @returns {string} PersonStatus
+   * @memberof PeopleSDKAdapter
+   */
+  getStatus(status) {
+    return Object.keys(PersonStatus).find((key) => PersonStatus[key] === status);
+  }
+
+  /**
+   * Encodes the personID into base64 format and return the UUID
```
adamweeks: If I'm reading the code correctly, this is actually decoding the base64 string then grabbing the last section. Probably would help to have some samples in the jsdoc.

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,119 @@
...
+   *
+   * @param {string} ID
+   * @returns {string}
+   * @memberof PeopleSDKAdapter
+   */
+  getPersonUUID(ID) {
+    return Buffer.from(ID, 'base64')
```
adamweeks: Can you use the helper method from the sdk? This uses the cisco secure requirement of the `UrlSafeBase64` decoder.

```js
import {base64, retry} from '@webex/common';

base64.decode(ID)
```

lalli-flores:
> Let's standardize on package updates commit messages being:
> `fix(package): update webex dependency` instead of `build(webex): update dependency`

On this, let's say what version we are updating to, if an update
`fix(package): update webex to version XXX`

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,119 @@
...
+import {PeopleAdapter, PersonStatus} from '@webex/component-adapter-interfaces';
+import {publish, refCount, flatMap, map, filter} from 'rxjs/operators';
+
+const USER_PRESENCE_UPDATE_EVENT = 'event:apheleia.subscription_update';
+
+/**
+ * The `PeopleSDKAdapter` is an implementation of the `PeopleAdapter` interface.
+ * This adapter utilizes the Webex JS SDK to fetch data about a person.
+ *
+ * @export
```
lalli-flores: Is this standard? I couldn't find this tag in the of the JS docs

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,119 @@
...
+  constructor(datasource) {
+    super(datasource);
+
+    this.getPersonObservables = {};
+  }
+
+  /**
+   * Returns a proper Enum key for a given PersonStatus value
+   *
+   * @param {string} status
```
lalli-flores: Can you add the private tag to indicate this is intended to be a helper method? See [@private](https://jsdoc.app/tags-private.html)
```suggestion
   * @private
   * @param {string} status
```

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,119 @@
...
+   * @memberof PeopleSDKAdapter
+   */
+  getStatus(status) {
+    return Object.keys(PersonStatus).find((key) => PersonStatus[key] === status);
+  }
+
+  /**
+   * Encodes the personID into base64 format and return the UUID
+   *
+   * @param {string} ID
```
lalli-flores: Can you add the private tag to indicate this is intended to be a helper method? See [@private](https://jsdoc.app/tags-private.html)
```suggestion
   * @private
   * @param {string} ID
```

src/PeopleSDKAdapter.js
lalli-flores: I think it's documenting strings should keep the implementation details out of the description (as you can get that by reading the code), and instead focus on the result.
```suggestion
   * Returns the UUID of a person from the given public ID
```
Not super sure on the subtleties of UUID vs hydra ID but i hope you get the idea.

As for the examples, if we are going to do that, let's do it only on the "public" ones

src/PeopleSDKAdapter.js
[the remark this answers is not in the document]
lalli-flores: What "shape" are you referring to?
We should have descriptions on what we expect the methods to take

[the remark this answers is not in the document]
lalli-flores: Also, we should add the private tag to indicate this is intended to be a helper method? See [@private](https://jsdoc.app/tags-private.html)
```suggestion
   * @private
   * @param {string} ID  ID of the person for which to fetch data
```

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,119 @@
...
+  /**
+   * Returns an observable that emits person data of the given ID.
+   *
+   * @param {string} ID ID of person to get.
+   * @returns {Observable.<Person>}
+   * @memberof PeopleSDKAdapter
+   */
+  getPerson(ID) {
+    if (!(ID in this.getPersonObservables)) {
+      const personObs = from(this.fetchPerson(ID)).pipe(
```
lalli-flores: I don't think the mapping and error are necessary, as those should be handled by the observable transformation already

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,119 @@
...
+        map((data) => data),
+        (error) => error
+      );
+      const presenceObs = from(this.datasource.internal.presence.subscribe(this.getPersonUUID(ID))).pipe(
+        map((data) => data.responses[0].status.status, (error) => error)
+      );
+
+      // Subscribe to our internal `Apheleia` server to listen for any status change
```
lalli-flores:
```suggestion
      // Subscribe to `Apheleia` internal service to listen for presence changes
```

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,119 @@
...
+   * @memberof PeopleSDKAdapter
+   */
+  getPerson(ID) {
+    if (!(ID in this.getPersonObservables)) {
+      const personObs = from(this.fetchPerson(ID)).pipe(
+        map((data) => data),
+        (error) => error
+      );
+      const presenceObs = from(this.datasource.internal.presence.subscribe(this.getPersonUUID(ID))).pipe(
+        map((data) => data.responses[0].status.status, (error) => error)
```
lalli-flores: What does `(error) => error` do? is it necessary?

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,119 @@
...
+
+      // Teardown logic
+      const teardownObs = Observable.create(() => async () => {
+        await this.datasource.internal.presence.unsubscribe(this.getPersonUUID(ID));
+      });
+
+      // fetch the original person data on the first run
+      // and updated object after each status change
+      // include the teardown logic to our finalized observable
+      const source = merge(personStatusObs, UpdatedPersonObs, teardownObs);
```
lalli-flores: `merge` is not the right operator here. If you look at the [marble diagram](http://reactivex.io/documentation/operators/merge.html), this operator combines the output of the observables as they come in, and not necessarily in the order you place them but that's not quite what as we cannot guarantee when an observable will complete.

We want to use [`concat`](https://rxjs-dev.firebaseapp.com/api/index/function/concat) instead, which guarantees order by waiting for the first observable to complete before going to the next.

So the `source` observable should look like:
```
const source = concat(personObs, statusUpdateObs);
```

Where the first `personObs` is created using [`fromPromise`](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#static-method-fromPromise) specifically so it completes.

Also, I don't think the `teardown` logic is something we want in an observable, but maybe in an operator like [`finalize`](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-finalize)

src/PeopleSDKAdapter.js
akoushke: then if that's the case, `webex` module needs to be a dependency. not a `peer` anymore.

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,119 @@
...
+        (error) => error
+      );
+      const presenceObs = from(this.datasource.internal.presence.subscribe(this.getPersonUUID(ID))).pipe(
+        map((data) => data.responses[0].status.status, (error) => error)
+      );
+
+      // Subscribe to our internal `Apheleia` server to listen for any status change
+      // And update the person object with status response from the subscription
+      const personStatusObs = presenceObs.pipe(
+        flatMap((status) => personObs.pipe(map((person) => ({...person, status: this.getStatus(status)}))))
```
lalli-flores: This will still make another network request, right?

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,119 @@
...
+        map((event) => this.getStatus(event.data.status))
+      );
+
+      // Update the person status after each change emitted from the event
+      const UpdatedPersonObs = personObs.pipe(
+        flatMap((person) => eventObs.pipe(map((status) => ({...person, status}))))
+      );
+
+      // Teardown logic
+      const teardownObs = Observable.create(() => async () => {
```
lalli-flores: Creation callbacks should not be `async` because we can't guarantee that they will resolve or when they will resolve

src/PeopleSDKAdapter.js
akoushke: not catching the error will crash the application.

src/PeopleSDKAdapter.test.js
```diff
@@ -0,0 +1,115 @@
...
+      peopleSDKAdapter.getPerson('id').subscribe(
+        () => {},
+        (error) => {
+          expect(error.message).toBe(errorMessage);
+          done();
+        }
+      );
+    });
+  });
+
```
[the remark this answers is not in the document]
akoushke: no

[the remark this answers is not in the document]
akoushke: all methods are covered and tested after fetching a mock person

src/PeopleSDKAdapter.js
akoushke: no

src/PeopleSDKAdapter.js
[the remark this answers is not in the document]
akoushke: take a look [here](https://github.com/webex/sdk-component-adapter/blob/master/src/RoomsSDKAdapter.js#L25)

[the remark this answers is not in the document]
akoushke: the `private` tag is not included into other module in this project. I don't think we should just add it now.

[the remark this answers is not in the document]
lalli-flores: The only other file on this repo that has private methods is `RoomsSDKAdapter` and we will be modifying it and I will ask to add the same `@private` tag for private methods for consistency within this repo.
We already use this on the [JS SDK](https://github.com/webex/webex-js-sdk/blob/d01be55fb21d7818a5512fb28d73b920daedec93/packages/node_modules/%40webex/webex-core/src/webex-core.js) so we're not inconsistent with other repos.

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,112 @@
...
+   * @memberof PeopleSDKAdapter
+   */
+  getPerson(ID) {
+    if (!(ID in this.getPersonObservables)) {
+      const personObs = from(this.fetchPerson(ID));
+      const presenceObs = from(this.datasource.internal.presence.subscribe(this.getPersonUUID(ID))).pipe(
+        map((data) => data.responses[0].status.status)
+      );
+
+      // Subscribe to `Apheleia` internal service to listen for presence changes
```
lalli-flores: Shouldn't this comment be before line 75?

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,112 @@
...
+      // And update the person object with status response from the subscription
+      const personStatusObs = presenceObs.pipe(
+        flatMap((status) => personObs.pipe(map((person) => ({...person, status: this.getStatus(status)}))))
+      );
+
+      // Listen to any status change for the given person ID
```
lalli-flores:
```suggestion
      // Listen for presence changes for the given person ID after subscription to service
```

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,112 @@
...
+  getPerson(ID) {
+    if (!(ID in this.getPersonObservables)) {
+      const personObs = from(this.fetchPerson(ID));
+      const presenceObs = from(this.datasource.internal.presence.subscribe(this.getPersonUUID(ID))).pipe(
+        map((data) => data.responses[0].status.status)
+      );
+
+      // Subscribe to `Apheleia` internal service to listen for presence changes
+      // And update the person object with status response from the subscription
+      const personStatusObs = presenceObs.pipe(
```
lalli-flores: Since nothing else is using `presenceObs`, can you just combine the pipes into one? e.g.
```
const personStatus = from(this.datasource.internal.presence.subscribe(this.getPersonUUID(ID))).pipe(
        map((data) => data.responses[0].status.status)
        flatMap((status) => person.pipe(map((data) => ({...data, status: this.getStatus(status)}))))
      );

File: src/PeopleSDKAdapter.js [hunk 24]
```diff
@@ -0,0 +1,112 @@
...
+  /**
+   * Returns an observable that emits person data of the given ID.
+   *
+   * @param {string} ID ID of person to get.
+   * @returns {Observable.<Person>}
+   * @memberof PeopleSDKAdapter
+   */
+  getPerson(ID) {
+    if (!(ID in this.getPersonObservables)) {
+      const personObs = from(this.fetchPerson(ID));
```
lalli-flores: Can we just call this `person`? Not sure I personally like the `obs` prefix, i think it makes the readability flow a little harder (See my other comment on line 81

Maybe adopt [Angular's naming convention](https://angular.io/guide/rx-library#naming-conventions-for-observables)?

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,112 @@
...
+      );
+
+      // Subscribe to `Apheleia` internal service to listen for presence changes
+      // And update the person object with status response from the subscription
+      const personStatusObs = presenceObs.pipe(
+        flatMap((status) => personObs.pipe(map((person) => ({...person, status: this.getStatus(status)}))))
+      );
+
+      // Listen to any status change for the given person ID
+      const eventObs = fromEvent(this.datasource.internal.mercury, USER_PRESENCE_UPDATE_EVENT).pipe(
```
lalli-flores: Can you call this `presenceUpdate`?

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,112 @@
...
+        filter((event) => event.data.subject === this.getPersonUUID(ID)),
+        map((event) => this.getStatus(event.data.status))
+      );
+
+      // Update the person status after each change emitted from the event
+      const UpdatedPersonObs = personObs.pipe(
```
lalli-flores: Can you call this `personUpdate`? This way the naming is consistent on the event observables

src/PeopleSDKAdapter.test.js
```diff
@@ -0,0 +1,111 @@
...
+      lastName: 'Components',
+      avatar: 'avatar',
+      orgId: 'orgID',
+      created: 'created',
+      type: 'person',
+    };
+    personID = 'personID';
+  });
+
+  test('fetchPerson() fetches person details properly', async () => {
```
lalli-flores:
```suggestion
  test('fetchPerson() returns Person object', async () => {
```
or alternatively,
```suggestion
  test('fetchPerson() returns a person in a proper shape', async () => {
```

src/PeopleSDKAdapter.test.js
```diff
@@ -0,0 +1,111 @@
...
+    });
+
+    test('stops listening to events when unsubscribing', () => {
+      peopleSDKAdapter.fetchPerson = jest.fn(() => Promise.resolve(mockSDKPerson));
+      mockSDK.internal.presence.subscribe = jest.fn(() => Promise.resolve({responses: [{status: {status: 'active'}}]}));
+
+      const subscription = peopleSDKAdapter.getPerson(personID).subscribe();
+
+      subscription.unsubscribe();
+      expect(mockSDK.internal.presence.unsubscribe).toHaveBeenCalled();
```
lalli-flores: 👍

src/PeopleSDKAdapter.test.js
```diff
@@ -0,0 +1,111 @@
...
+    });
+
+    test('throws a proper error message for an invalid person ID to fetch the person details from', (done) => {
```
lalli-flores:
```suggestion
    test('throws error for an invalid person ID', (done) => {
```

src/PeopleSDKAdapter.test.js
```diff
@@ -0,0 +1,111 @@
...
+      peopleSDKAdapter.getPerson(personID).subscribe(
+        () => {},
+        (error) => {
+          expect(error.message).toBe(errorMessage);
+          done();
+        }
+      );
+    });
+
+    test("throws a proper error message for an invalid person ID to subscribe to person's presence status update", (done) => {
```
lalli-flores: Not sure why this one use `"` instead of `'`
```suggestion
    test('throws error on failed presence update subscription', (done) => {
```

src/PeopleSDKAdapter.test.js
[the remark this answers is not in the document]
lalli-flores: If we are going to add unit tests for public methods only that's fine, but why add tests for `fetchPerson` only then?

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,112 @@
...
+      // fetch the original person data on the first run
+      // and updated object after each status change
+      // include the teardown logic to our finalized observable
+      const source = concat(personStatusObs, UpdatedPersonObs);
+
+      this.getPersonObservables[ID] = source.pipe(
+        publish(),
+        refCount(),
+        finalize(() => {
+          this.datasource.internal.presence.unsubscribe(this.getPersonUUID(ID));
```
lalli-flores: Don't forget to delete the observable from `getPersonObservables`

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,112 @@
...
+
+  /**
+   * Returns the UUID of a person from the given public ID
+   *
+   * @param {string} ID
+   * @returns {string}
+   * @memberof PeopleSDKAdapter
+   */
+  getPersonUUID(ID) {
+    return base64.decode(ID).split('/')[4]; // ciscospark://us/PEOPLE/<UUID>
```
adamweeks: Sorry, I totally forgot we are already doing this work with helpers in webex core:
https://github.com/webex/webex-js-sdk/blob/master/packages/node_modules/@webex/common/src/uuid-utils.js#L45

src/PeopleSDKAdapter.js
adamweeks: I like the angular naming convention of putting the `$` at the end. I was wondering about that when reading through all the docs and example code.

src/PeopleSDKAdapter.js
adamweeks: I find it more readable to split them out, but I think `presenceObs` could be `presenceSubscription$`

src/PeopleSDKAdapter.js
adamweeks: @lalli-flores should it be `presence$`?

src/PeopleSDKAdapter.js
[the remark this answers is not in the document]
akoushke: I'm talking about `webex/component` repo.

src/PeopleSDKAdapter.test.js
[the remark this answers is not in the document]
akoushke: `fetchPerson` unit test had been removed.

src/PeopleSDKAdapter.test.js
```diff
@@ -0,0 +1,82 @@
...
+    personID = 'personID';
+  });
+
+  describe('getPerson()', () => {
+    test('returns an observable', () => {
+      expect(isObservable(peopleSDKAdapter.getPerson(personID))).toBeTruthy();
+    });
+
+    test('returns a person in a proper shape', (done) => {
+      mockSDK.internal.presence.subscribe = jest.fn(() => Promise.resolve({responses: [{status: {status: 'active'}}]}));
```
lalli-flores: Isn't this already done in the `mockSDK`?

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,120 @@
+import {from, fromEvent, concat} from 'rxjs';
+import {deconstructHydraId} from '@webex/common';
+import {PeopleAdapter, PersonStatus} from '@webex/component-adapter-interfaces';
+import {flatMap, filter, finalize, map, publishReplay, refCount} from 'rxjs/operators';
```
adamweeks: Should we put all the rxjs imports together? Also, can you alphabetize?

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,120 @@
+import {from, fromEvent, concat} from 'rxjs';
```
adamweeks: Can you alphabetize these?

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,120 @@
...
+   *
+   * @public
+   * @param {string} ID ID of person to get.
+   * @returns {Observable.<Person>}
+   * @memberof PeopleSDKAdapter
+   */
+  getPerson(ID) {
+    if (!(ID in this.getPersonObservables)) {
+      const person$ = from(this.fetchPerson(ID));
+
```
adamweeks: Could you store the UUID here and use the variable instead of running `this.getPersonUUID` every time? 

```js
const personUUID = this.getPersonUUID(ID);
```

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,120 @@
...
+      const personUpdate$ = person$.pipe(
+        flatMap((person) => presenceUpdate$.pipe(map((status) => ({...person, status}))))
+      );
+
+      // Fetch the original person data on the first run
+      // and updated object after each status change
+      // include the teardown logic to our finalized observable.
+      // Unsubscribe from `Apheleia` internal service, when
+      // all the subscriptions have been completed.
+      const source = concat(personStatus$, personUpdate$).pipe(
```
adamweeks: For the rooms adapter, I went with the naming convention of calling this the name of the function to show it is the final observable:
```suggestion
      const getPerson$ = concat(personStatus$, personUpdate$).pipe(
```

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,120 @@
...
+    if (!(ID in this.getPersonObservables)) {
+      const person$ = from(this.fetchPerson(ID));
+
+      // Subscribe to `Apheleia` internal service to listen for presence changes
+      // And update the person object with status response from the subscription
+      const presenceSubscription$ = from(this.datasource.internal.presence.subscribe(this.getPersonUUID(ID))).pipe(
+        map((data) => data.responses[0].status.status)
+      );
+      const personStatus$ = presenceSubscription$.pipe(
+        flatMap((status) => person$.pipe(map((person) => ({...person, status: this.getStatus(status)}))))
```
adamweeks: Do we have to fetch the person data (`person$`) for every update or can we store the result of the initial fetch and just modify it with the changed status?

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,120 @@
...
+  /**
+   * Returns the UUID of a person from the given public ID
+   *
+   * @private
+   * @param {string} ID
+   * @returns {string}
+   * @memberof PeopleSDKAdapter
+   */
+  getPersonUUID(ID) {
+    return deconstructHydraId(ID).id; // ciscospark://us/PEOPLE/<UUID>
```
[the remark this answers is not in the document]
akoushke: yes method had been invoked in several places and we are mapping the data inside this method to return only `ID`

src/PeopleSDKAdapter.js
akoushke: we are not fetching. `from` stores the data for you automatically. Unless you invoke the method/from operator again.

src/PeopleSDKAdapter.js
akoushke: hella smarttt

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,107 @@
...
+    super(datasource);
+
+    this.getPersonObservables = {};
+  }
+
+  /**
+   * Returns a proper Enum key for a given PersonStatus value
+   *
+   * @private
+   * @param {string} status
```
lalli-flores:
```suggestion
   * @param {string} status  Person status from Apheleia service.
```

src/PeopleSDKAdapter.js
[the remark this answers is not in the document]
lalli-flores: > Can you still add the argument description?

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,107 @@
...
+
+      // Subscribe to `Apheleia` internal service to listen for presence changes
+      // And update the person object with status response from the subscription
+      const personWithStatus$ = from(this.datasource.internal.presence.subscribe(personUUID)).pipe(
+        map((data) => data.responses[0].status.status),
+        flatMap((status) => person$.pipe(map((person) => ({...person, status: this.getStatus(status)}))))
+      );
+
+      // Listen for presence changes for the given person ID after subscription to service
+      const presenceUpdate$ = fromEvent(this.datasource.internal.mercury, USER_PRESENCE_UPDATE_EVENT).pipe(
```
lalli-flores: Can we be consistent with the wording? Seems like `presence` and `status` are being used interchangeably, can we just stick with one?
My suggestion would be `status` since that's what the person object expects, and that's what we're calling it everywhere else.
This would then be `statusUpdate$`, and comment should be updated accordingly.

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,107 @@
...
+   * @param {string} ID ID of person to get.
+   * @returns {Observable.<Person>}
+   * @memberof PeopleSDKAdapter
+   */
+  getPerson(ID) {
+    if (!(ID in this.getPersonObservables)) {
+      const personUUID = deconstructHydraId(ID).id;
+      const person$ = from(this.fetchPerson(ID));
+
+      // Subscribe to `Apheleia` internal service to listen for presence changes
```
lalli-flores: Let's keep wording consistent
```suggestion
      // Subscribe to `Apheleia` internal service to listen for status changes
```

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,107 @@
...
+      const personUpdate$ = person$.pipe(
+        flatMap((person) => presenceUpdate$.pipe(map((status) => ({...person, status}))))
+      );
+
+      // Fetch the original person data on the first run
+      // and updated object after each status change
+      const getPerson$ = concat(personWithStatus$, personUpdate$).pipe(
+        finalize(() => {
+          // Unsubscribe from `Apheleia` internal service
+          // and remove the targeted connectedObservable
```
lalli-flores: Can you remove this line? It's misleading (and untrue) as there are no connected observables at this point.
Comment still flows without this

lalli-flores:
This is missing integration tests. We should merge https://github.com/webex/sdk-component-adapter/pull/10 before merging this, but the tests can be written

src/PeopleSDKAdapter.js
akoushke: it is true but yeah sure.

lalli-flores: There are no connected observables here, it's just a regular observable. A `ConnectedObservable` only gets created by `publishReplay`, all other operators return a regular `Observable`.

akoushke: okay sure lalli. Is there anything else?

akoushke:
> This is missing integration tests. We should merge #10 before merging this, but the tests can be written

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-105443 ticket had been created.

src/PeopleSDKAdapter.js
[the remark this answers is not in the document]
akoushke: I already did!

src/PeopleSDKAdapter.js
```diff
@@ -0,0 +1,106 @@
...
+      const person$ = from(this.fetchPerson(ID));
+
+      // Subscribe to `Apheleia` internal service to listen for status changes
+      // And update the person object with status response from the subscription
+      const personWithStatus$ = from(this.datasource.internal.presence.subscribe(personUUID)).pipe(
+        map((data) => data.responses[0].status.status),
+        flatMap((status) => person$.pipe(map((person) => ({...person, status: this.getStatus(status)}))))
+      );
+
+      // Listen for presence changes for the given person ID after subscription to service
```
lalli-flores:
```suggestion
      // Listen for status changes for the given person ID after subscription to service
```

PR #19, opened 2019-11-13
refactor(RoomsSdkAdapter): convert to observable stream

adamweeks:


src/RoomsSDKAdapter.js
```diff
@@ -36,42 +37,67 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+   * @memberof RoomsSDKAdapter
+   */
+  addRoomChangeListener(ID) {
+    const roomChange$ = fromEvent(this.datasource.rooms, ROOM_UPDATED_EVENT);
+
+    return roomChange$.pipe(
+      // Is the room change event for our subscribed room?
+      filter((event) => event.data.id === ID),
+      // Event data doesn't have the room data in it, so we need to fetch manually
+      mergeMap(async () => this.fetchRoom(ID))
```
lalli-flores: Why are you using `mergeMap` with an `async` callback here, instead of using `from` to convert the promise to observable?

src/RoomsSDKAdapter.js
```diff
@@ -36,42 +37,67 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
-        // Start listening for room changes
-        this.datasource.rooms.listen();
-        this.datasource.rooms.on(ROOM_UPDATED_EVENT, () => {
-          // Room has updates, fetch and send
-          this.fetchRoom(ID)
-            .then((room) => observer.next(room))
-            .catch((error) => observer.error(error));
-        });
+      // Tell the sdk to start listening to room changes
+      this.datasource.rooms.listen();
```
lalli-flores: Shouldn't this be done at the adapter level?

src/RoomsSDKAdapter.js
```diff
@@ -36,42 +37,67 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+   * Create an observable that subscribes to room change events emitted via websocket and
+   * emits the changed room object.
+   *
+   * @private
+   * @param {string} ID ID of room to get
+   * @returns {Observable.<Room>} Observable that emits changes to the room
+   * @memberof RoomsSDKAdapter
+   */
+  addRoomChangeListener(ID) {
+    const roomChange$ = fromEvent(this.datasource.rooms, ROOM_UPDATED_EVENT);
```
lalli-flores: We should standardize on the wording, `change` vs `update`. Personally, I prefer `update`

src/RoomsSDKAdapter.js
```diff
@@ -36,42 +37,67 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
-          .catch((error) => observer.error(error));
-
-        return () => {
-          // Cleanup when subscription count is 0
-          this.datasource.rooms.stopListening();
-          this.datasource.rooms.off(ROOM_UPDATED_EVENT);
-          delete this.getRoomObservables[ID];
-        };
-      });
+      const getRoom$ = this.getRoomObservable(ID);
```
lalli-flores: I appreciate trying to make the code more modular, but let's be careful of over-engineering. I think the logic for this observable is so complex that needs to be broken down in 3 methods. See my [sample attempt](https://github.com/lalli-flores/sdk-component-adapter/blob/rooms-remove-async/src/RoomsSDKAdapter.js#L63)

adamweeks: There's so much logic in the `getRoomObservable` function, it is easier to comprehend by splitting it into its own function.

akoushke: I see the that you take the needed extra work to a different method, where I just left mine in the same method. We just need to come up with a conclusion on how we want to implement this. I personally don't care *much*, though I'm leaning a bit towards my approach to make the code looks simpler and easier to read and have all the observables together to look for.

akoushke: but whatever pattern we are coming up with, needs to be applied in both PRs.

src/RoomsSDKAdapter.js
adamweeks: What do you mean? This is the rooms adapter.

akoushke:
Screen Shot 2019-11-13 at 2 59 42 PM
what do you think Adam?

src/RoomsSDKAdapter.js
```diff
@@ -36,41 +39,82 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+  }
+
+  /**
+   * Tells the SDK to stop listening to room events.
+   *
+   * Note: Since the SDK listens to ALL room events, this function only
+   * calls the SDK's `rooms.stopListening` function once all of the listeners are done.
+   * If `rooms.stopListening` is called early, existing subscribers won't get any updates.
+   *
+   * @memberof RoomsSDKAdapter
```
lalli-flores:
```suggestion
   * @private
   * @memberof RoomsSDKAdapter
```

src/RoomsSDKAdapter.js
```diff
@@ -17,11 +17,14 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
     this.getRoomObservables = {};
+    this.isListening = false;
+    this.listenerCount = 0;
   }
 
   /**
    * Fetches the room data from the sdk and returns in the shape required by adapter.
    *
+   * @private
    * @param {string} ID
```
lalli-flores:
```suggestion
   * @param {string} ID  ID of the room for which to fetch data
```

src/RoomsSDKAdapter.test.js
```diff
@@ -16,6 +16,8 @@ describe('Rooms SDK Adapter', () => {
       type: 'group',
       title: 'mock room',
     };
+
+    mockSDK.rooms.get = jest.fn(() => Promise.resolve(mockSDKRoom));
```
lalli-flores: Is there a reason why this is not defined within the mock and done here?
`mockSDKRoom` could be exported if necessary (i think only one test would need it if removed from here)

src/RoomsSDKAdapter.js
```diff
@@ -36,41 +39,82 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+
+      // The observable flow for fetching room data, then listening for websocket events about room changes.
+      const getRoom$ = concat(
+        // Fetch Our Room Data
+        room$,
+        roomUpdate$
+      ).pipe(
+        finalize(() => {
+          // Called once all subscriptions to `ID` are done.
           delete this.getRoomObservables[ID];
```
akoushke: can we switch these two lines? 🙈

src/RoomsSDKAdapter.js
```diff
@@ -36,41 +39,82 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+   *
+   * Note: Since the SDK listens to ALL room events, this function only
+   * calls the SDK's `rooms.listen` function on the first room to listen.
+   * Repeated calls to `rooms.listen` are not needed afterwards.
+   *
+   * @private
+   * @memberof RoomsSDKAdapter
+   */
+  startListeningToRoomUpdates() {
+    if (!this.isListening) {
```
lalli-flores: Can't you use `listenerCount` and remove `isListening`?
```suggestion
    if (this.listenerCount > 0) {
```

adamweeks: Yep. Leftover artifact from testing code.

src/RoomsSDKAdapter.js
```diff
@@ -1,5 +1,5 @@
-import {Observable} from 'rxjs';
-import {publish, refCount} from 'rxjs/operators';
+import {fromEvent, from, concat} from 'rxjs';
+import {filter, finalize, publishReplay, refCount, flatMap} from 'rxjs/operators';
```
akoushke: can we order this alphabetically please?

akoushke:
sorry one more thing, can you make sure to nullify all the variables inside `afterEach` and mabye combine `let` statements together? `let a,b,c` ?

PR #20, opened 2019-11-14
@webex/common custom export - rollup

akoushke:
```js
Object.defineProperty(exports, 'deconstructHydraId', {
  enumerable: true,
  get: function get() {
    return _uuidUtils.deconstructHydraId;
  }
});
```
will force us to use `custom exports` for the methods to be appropriately exported. 

Also, I turned on the `Prod` env to build.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -34,7 +34,11 @@ export default [
       babel({
         runtimeHelpers: true,
       }),
-      commonJS(),
+      commonJS({
+        // TODO: Remove workaround once fixed in SDK
+        // explicitly specify unresolvable named exports
+        namedExports: {'@webex/common': ['deconstructHydraId']},
+      }),
       json(),
       builtins(),
     ],


package.json
```diff
@@ -4,7 +4,7 @@
   "main": "dist/webexSDKComponentAdapter.cjs.js",
   "module": "dist/webexSDKComponentAdapter.esm.js",
   "scripts": {
-    "build": "rollup -c",
+    "build": "rollup -c --environment INCLUDE_DEPS,BUILD:production",
```
lalli-flores: Can you explain the environment variables you just added and what benefits they have?

rollup.config.js
```diff
@@ -34,7 +34,10 @@ export default [
       babel({
         runtimeHelpers: true,
       }),
-      commonJS(),
+      commonJS({
+        // explicitly specify unresolvable named exports
+        namedExports: {'@webex/common': ['deconstructHydraId']},
```
lalli-flores: Can you add a TODO comment to remove this workaround once fixed on the SDK?

package.json
akoushke: don't feel like it hehe. its removed anyways.

PR #21, opened 2019-11-18
SDK Adapter connect/disconnect

akoushke:
Implement `connect/disconnect` inside `webexSDKAdapter` and apply the changes where necessary.

### DIFF
--- a/src/RoomsSDKAdapter.integration.test.js
+++ b/src/RoomsSDKAdapter.integration.test.js
@@ -1,8 +1,9 @@
 import {createIntegrationTestUser, removeIntegrationTestUser} from './testHelper';
-import RoomsSDKAdapter from './RoomsSDKAdapter';
+
+import WebexSDKAdapter from './';
 
 describe('Rooms SDK Adapter', () => {
-  let roomsSDKAdapter, subscription, user;
+  let createdRoom, getRoom$, webexSDKAdapter, subscription, user;
 
   // Since these are integration tests with live data,
   // increase the async "idle" timeout so jest doesn't error early.
@@ -10,33 +11,40 @@ describe('Rooms SDK Adapter', () => {
 
   beforeAll(async () => {
     user = await createIntegrationTestUser();
-  });
-
-  beforeEach(() => {
-    roomsSDKAdapter = new RoomsSDKAdapter(user.sdk);
+    webexSDKAdapter = new WebexSDKAdapter(user.sdk);
+    await webexSDKAdapter.connect();
   });
 
   afterAll(async () => {
-    // This is done after unsubscribing, but since unsubscribe can't be async, manually call it.
-    await user.sdk.rooms.stopListening();
-
     try {
       await removeIntegrationTestUser(user);
+      await webexSDKAdapter.disconnect();
     } catch (reason) {
       // eslint-disable-next-line no-console
-      console.warn('Failed to delete test user for Room SDK Adapter integration tests', reason);
+      console.warn('Failed to delete test user for Room SDK Adapter integration tests.', reason);
     }
   });
 
   describe('getRoom() returns', () => {
-    afterEach(() => {
-      subscription.unsubscribe();
+    beforeEach(async () => {
+      createdRoom = await user.sdk.rooms.create({title: 'Webex Teams Test Room'});
+      getRoom$ = webexSDKAdapter.roomsAdapter.getRoom(createdRoom.id);
+    });
+
+    afterEach(async () => {
+      try {
+        await user.sdk.rooms.remove(createdRoom.id);
+        subscription.unsubscribe();
+        getRoom$ = null;
+        createdRoom = null;
+      } catch (reason) {
+        // eslint-disable-next-line no-console
+        console.warn(`Failed to delete a room "${createdRoom.id}" in Room SDK Adapter integration tests.`, reason);
+      }
     });
 
     test('a room in a proper shape', async (done) => {
-      const createdRoom = await user.sdk.rooms.create({title: 'Webex Teams Test Room'});
-
-      subscription = roomsSDKAdapter.getRoom(createdRoom.id).subscribe((room) => {
+      subscription = getRoom$.subscribe((room) => {
         expect(room).toMatchObject({
           ID: createdRoom.id,
           title: createdRoom.title,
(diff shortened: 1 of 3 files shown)


src/RoomsSDKAdapter.integration.test.js
```diff
@@ -1,42 +1,44 @@
...
+      await webexSDKAdapter.disconnect();
     } catch (reason) {
       // eslint-disable-next-line no-console
       console.warn('Failed to delete test user for Room SDK Adapter integration tests', reason);
     }
   });
 
   describe('getRoom() returns', () => {
+    beforeEach(async () => {
+      createdRoom = await user.sdk.rooms.create({title: 'Webex Teams Test Room'});
```
lalli-flores: I don't think it's necessary to create a new room per test, I think one room for this suite should be sufficient

src/RoomsSDKAdapter.integration.test.js
```diff
@@ -1,42 +1,44 @@
...
+      getRoom$ = webexSDKAdapter.roomsAdapter.getRoom(createdRoom.id);
+    });
+
     afterEach(() => {
+      createdRoom = null;
```
lalli-flores: Should we actually delete from the platform? This should also happen at the end of the suite run

src/RoomsSDKAdapter.integration.test.js
```diff
@@ -1,42 +1,44 @@
...
     } catch (reason) {
       // eslint-disable-next-line no-console
       console.warn('Failed to delete test user for Room SDK Adapter integration tests', reason);
     }
   });
 
   describe('getRoom() returns', () => {
+    beforeEach(async () => {
+      createdRoom = await user.sdk.rooms.create({title: 'Webex Teams Test Room'});
+      getRoom$ = webexSDKAdapter.roomsAdapter.getRoom(createdRoom.id);
```
lalli-flores: Can't you define this once at the beginning of the test run? The room ID wouldn't change if we have one room per run

src/WebexSDKAdapter.js
```diff
@@ -12,5 +12,23 @@ export default class WebexSDKAdapter {
   constructor(sdk) {
     this.peopleAdapter = new PeopleSDKAdapter(sdk);
     this.roomsAdapter = new RoomsSDKAdapter(sdk);
+    this.sdk = sdk;
+  }
+
+  /**
+   * Call to register the device to the cloud and start the
```
lalli-flores: I think we should make this description a little more abstract, in case we do more things with method in the future.
Here's my -more generic- suggestion:
```
Connect to Webex services so SDK can listen for data updates.

File: src/WebexSDKAdapter.js [hunk 5]
```diff
@@ -12,5 +12,23 @@ export default class WebexSDKAdapter {
...
+   * mercury connection to utilize the internal services
+   */
+  async connect() {
+    await this.sdk.internal.device.register();
+    await this.sdk.internal.mercury.connect();
+  }
+
+  /**
+   * Call to stop the mercury connection and unregister the device from the cloud
```
lalli-flores: I think we should make this description a little more abstract, in case we do more things with method in the future.
Here's my -more generic- suggestion:
```
Disconnect from Webex services, closing any connections SDK may have opened.

File: src/RoomsSDKAdapter.integration.test.js (hunk 3 above)
Comment (akoushke) [reply src/RoomsSDKAdapter.integration.test.js]: The rooms must be created for each test, since we are modifying the data in one of the test cases. However, I did make sure to remove the rooms after being tested.

File: src/RoomsSDKAdapter.integration.test.js (hunk 2 above)
Comment (akoushke) [reply src/RoomsSDKAdapter.integration.test.js]: not sure how to delete the object. I could either nullify or `undefined` them.

File: src/RoomsSDKAdapter.integration.test.js (hunk 1 above)
Comment (akoushke) [reply src/RoomsSDKAdapter.integration.test.js]: The rooms must be created for each test, since we are modifying the data in one of the test cases. However, I did make sure to remove the rooms after being tested.

File: src/RoomsSDKAdapter.integration.test.js [hunk 6]
```diff
@@ -1,42 +1,45 @@
...
-
-  beforeEach(() => {
-    roomsSDKAdapter = new RoomsSDKAdapter(user.sdk);
+    webexSDKAdapter = new WebexSDKAdapter(user.sdk);
+    await webexSDKAdapter.connect();
   });
 
   afterAll(async () => {
-    // This is done after unsubscribing, but since unsubscribe can't be async, manually call it.
-    await user.sdk.rooms.stopListening();
```
adamweeks: Can you run `npm run test:integration` to verify the tests exit gracefully with this removed? I had to add this in order to get jest to exit properly. Just running `npm test` isn't enough to verify.

src/RoomsSDKAdapter.integration.test.js
```diff
@@ -1,42 +1,45 @@
...
   describe('getRoom() returns', () => {
-    afterEach(() => {
+    beforeEach(async () => {
+      createdRoom = await user.sdk.rooms.create({title: 'Webex Teams Test Room'});
+      getRoom$ = webexSDKAdapter.roomsAdapter.getRoom(createdRoom.id);
+    });
+
+    afterEach(async () => {
+      await user.sdk.rooms.remove(createdRoom.id);
+      getRoom$ = null;
```
adamweeks: Still think that setting these to null is extraneous since the jest test runner should be doing this for us anyway.

src/RoomsSDKAdapter.integration.test.js
akoushke: i did.

src/RoomsSDKAdapter.integration.test.js
akoushke: sure. I can remove them both.

lalli-flores: Do you have documentation for this?
It can be extraneous in some cases, like this one but it might matter in others, like re-using `createdRoom` that may point to a removed room

lalli-flores: For the record, @akoushke and I agreed that we will `null` the variables, but will send a warming when something failed to clean up. This way we won't be masking any potential issues, but not crashing the test run either

src/RoomsSDKAdapter.integration.test.js
```diff
@@ -1,42 +1,50 @@
...
-      const createdRoom = await user.sdk.rooms.create({title: 'Webex Teams Test Room'});
+    afterEach(async () => {
+      try {
+        await user.sdk.rooms.remove(createdRoom.id);
+        subscription.unsubscribe();
+        getRoom$ = null;
+        createdRoom = null;
+      } catch (reason) {
+        // eslint-disable-next-line no-console
+        console.warn(`Failed to delete a room "${createdRoom.id}" in Room SDK Adapter integration tests`, reason);
```
```suggestion
        console.warn(`Failed to delete a room "${createdRoom.id}" in Room SDK Adapter integration tests.`, reason);
```

PR #22, opened 2019-11-19
SDK Adapter - PeopleSDKAdapter test integration

akoushke:
People test integration

### DIFF
--- a/src/PeopleSDKAdapter.integration.test.js
+++ b/src/PeopleSDKAdapter.integration.test.js
@@ -1,0 +1,79 @@
+import {from} from 'rxjs';
+import {delayWhen, skip} from 'rxjs/operators';
+
+import {createIntegrationTestUser, removeIntegrationTestUser, getPersonHydraID} from './testHelper';
+
+import WebexSDKAdapter from './';
+
+describe('People SDK Adapter', () => {
+  let getPerson$, webexSDKAdapter, subscription, user, userID;
+
+  // Since these are integration tests with live data,
+  // increase the async "idle" timeout so jest doesn't error early.
+  jest.setTimeout(30000);
+  beforeAll(async () => {
+    user = await createIntegrationTestUser();
+    userID = getPersonHydraID(user.id);
+    webexSDKAdapter = new WebexSDKAdapter(user.sdk);
+    await webexSDKAdapter.connect();
+  });
+
+  afterAll(async () => {
+    try {
+      await removeIntegrationTestUser(user);
+      await webexSDKAdapter.disconnect();
+    } catch (reason) {
+      // eslint-disable-next-line no-console
+      console.warn('Failed to delete test user for People SDK Adapter integration tests.', reason);
+    }
+  });
+
+  describe('getPerson() returns', () => {
+    beforeEach(() => {
+      getPerson$ = webexSDKAdapter.peopleAdapter.getPerson(userID);
+    });
+
+    afterEach(() => {
+      getPerson$ = null;
+      subscription.unsubscribe();
+    });
+
+    test('a person in a proper shape', (done) => {
+      subscription = getPerson$.subscribe((person) => {
+        expect(person).toMatchObject({
+          ID: userID,
+          emails: [user.emailAddress],
+          displayName: user.displayName,
+        });
+        done();
+      });
+    });
+
+    test('an updated person status after subscribing', (done) => {
+      subscription = getPerson$
+        .pipe(
+          delayWhen(() => from(user.sdk.internal.presence.setStatus('active', 1500))),
+          skip(1)
+        )
+        .subscribe((person) => {
+          expect(person.status).toEqual('ACTIVE');
+          done();
+        });
+    });
+
+    test('support for multiple subscriptions', (done) => {
+      subscription = getPerson$.subscribe();
+
+      const secondSubscription = getPerson$.subscribe((room) => {
+        expect(room).toMatchObject({
+          ID: userID,
+          emails: [user.emailAddress],
+          displayName: user.displayName,
+        });
+        done();
+      });
+
+      subscription.add(secondSubscription);
+    });
+  });
+});
--- a/src/RoomsSDKAdapter.integration.test.js
+++ b/src/RoomsSDKAdapter.integration.test.js
@@ -1,3 +1,6 @@
+import {from} from 'rxjs';
+import {delayWhen, skip} from 'rxjs/operators';
+
 import {createIntegrationTestUser, removeIntegrationTestUser} from './testHelper';
 
 import WebexSDKAdapter from './';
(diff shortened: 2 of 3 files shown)


src/PeopleSDKAdapter.integration.test.js
```diff
@@ -0,0 +1,79 @@
...
+        });
+        done();
+      });
+    });
+
+    test('an updated person status after subscribing', async (done) => {
+      let hasUpdated = false;
+
+      subscription = getPerson$.subscribe(async (person) => {
+        if (!hasUpdated) {
```
akoushke: I tried different operators to skip this check. However, with the way that we are combing the listening part with initial fetch section, we must invoke `getPerson$` and subscribe to it to get all the changes. Hopefully this combination won't cause any additional problems. Otherwise we need to separate the logic.

akoushke: so It does not have anything to do with `rxjs`. our logic is blocking us to remove this check, since we fetch and subscribe at the same time.

akoushke:
time estimate: 26 secs.

src/testHelper.js
```diff
@@ -62,9 +63,20 @@ export async function createIntegrationTestUser() {
...
   });
 }
+
+/**
+ * Constructs a Hydra ID for a given UUID.
+ *
+ * @param {String} UUID Person UUID
+ * @returns {string}
+ */
+export function getPersonHydraID(UUID) {
```
akoushke: first, I tried to add the `Hydra` ID to the object. However deserializing the user during removing the user threw some errors, since that property does not exist to begin with. Therefore, I decided to create a common method to use. I'm sure this method will be utilized for other suites as we implement more test cases.

src/testHelper.js
```diff
@@ -62,9 +63,20 @@ export async function createIntegrationTestUser() {
  * @param {Object} user created in #createIntegrationTestUser()
  * @returns {Promise}
  */
+
```
lalli-flores: Is there a reason between the new line between the JS Doc and the method definition?

src/testHelper.js
```diff
@@ -1,6 +1,7 @@
 import dotenv from 'dotenv';
 import {createTestUser, removeTestUser} from '@webex/test-users';
 import Webex from 'webex';
+import {constructHydraId} from '@webex/common/dist/uuid-utils';
```
lalli-flores: Are you importing like this because of the export issue on the SDK for `@webex/common`?

src/PeopleSDKAdapter.integration.test.js
```diff
@@ -0,0 +1,79 @@
...
+    await webexSDKAdapter.connect();
+  });
+
+  afterAll(async () => {
+    try {
+      await removeIntegrationTestUser(user);
+      await webexSDKAdapter.disconnect();
+    } catch (reason) {
+      // eslint-disable-next-line no-console
+      console.warn('Failed to delete test user for Room SDK Adapter integration tests.', reason);
```
lalli-flores:
```suggestion
      console.warn('Failed to delete test user for People SDK Adapter integration tests.', reason);
```

src/PeopleSDKAdapter.integration.test.js
lalli-flores: So here's a suggestion I think might work (but i haven't tested it). We can use the [`skip(1)`](https://rxjs-dev.firebaseapp.com/api/operators/skip) operator to skip the initial data emission. We can update the user after subscribing, which should trigger `getPerson` for testing our scenario. Code would look something like:
```js
subscription = getPerson$.pipe(skip(1)).subscribe((person) => {
  expect(person.status).toEqual('ACTIVE');
   done();
});

await user.sdk.internal.presence.setStatus('active', 1500); // Once this resolves, the subscription above will run
```

akoushke: I exactly tried this approach too. But we still need to subscribe to detect the changes. And skip will not allow us to subscribe to any changes to begin with.

src/testHelper.js
akoushke: I really don’t care lol

akoushke: Whatever you want

src/testHelper.js
akoushke: No what do you mean? That’s how we export them regularly. And also we not bundling this code.

src/testHelper.js
lalli-flores: Can you remove it? I think it might have been accidental

akoushke: Let me get home. Can you approve? I will address your comments.

src/testHelper.js
lalli-flores: To me a "regular" import looks like:
```suggestion
import {constructHydraId} from '@webex/common';
```
We don't directly import from files

akoushke: Oh shit. I didn’t even see that

src/PeopleSDKAdapter.integration.test.js
lalli-flores: You're right, we need to include the async code as part of the pipeline. Otherwise we won't know when the promise resolved. Using [`delayWhen](https://rxjs-dev.firebaseapp.com/api/operators/delayWhen) seems to work. It also makes the code more reliable since we're not mixing async code with observables.

```js
    test('an updated person status after subscribing', (done) => {
      subscription = getPerson$
        .pipe(
          delayWhen(() => from(user.sdk.internal.presence.setStatus('active', 1500))),
          skip(1)
        )
        .subscribe((person) => {
          expect(person.status).toEqual('ACTIVE');
          done();
        });
    });

File: src/testHelper.js (hunk 3 above)
Comment (lalli-flores) [reply src/testHelper.js]: Don't forget to remove this, please

File: src/PeopleSDKAdapter.integration.test.js [hunk 6]
```diff
@@ -0,0 +1,79 @@
...
+          delayWhen(() => from(user.sdk.internal.presence.setStatus('active', 1500))),
+          skip(1)
+        )
+        .subscribe((person) => {
+          expect(person.status).toEqual('ACTIVE');
+          done();
+        });
+    });
+
+    test('support for multiple subscriptions', async (done) => {
```
lalli-flores:
```suggestion
    test('support for multiple subscriptions' (done) => {
```

src/PeopleSDKAdapter.integration.test.js
```diff
@@ -0,0 +1,79 @@
...
+    beforeEach(async () => {
+      getPerson$ = webexSDKAdapter.peopleAdapter.getPerson(userID);
+    });
+
+    afterEach(async () => {
+      getPerson$ = null;
+      subscription.unsubscribe();
+    });
+
+    test('a person in a proper shape', async (done) => {
```
lalli-flores:
```suggestion
    test('a person in a proper shape', (done) => {
```

src/PeopleSDKAdapter.integration.test.js
```diff
@@ -0,0 +1,79 @@
...
+      await removeIntegrationTestUser(user);
+      await webexSDKAdapter.disconnect();
+    } catch (reason) {
+      // eslint-disable-next-line no-console
+      console.warn('Failed to delete test user for People SDK Adapter integration tests.', reason);
+    }
+  });
+
+  describe('getPerson() returns', () => {
+    beforeEach(async () => {
```
lalli-flores:
```suggestion
    beforeEach(() => {
```

src/PeopleSDKAdapter.integration.test.js
```diff
@@ -0,0 +1,79 @@
...
+      getPerson$ = webexSDKAdapter.peopleAdapter.getPerson(userID);
+    });
+
+    afterEach(async () => {
```
lalli-flores:
```suggestion
    afterEach(() => {
```

src/RoomsSDKAdapter.integration.test.js
```diff
@@ -54,21 +57,17 @@ describe('Rooms SDK Adapter', () => {
     });
 
     test('an updated room title after subscribing', async (done) => {
```
```suggestion
    test('an updated room title after subscribing', (done) => {
```

src/RoomsSDKAdapter.integration.test.js
```diff
@@ -54,21 +57,17 @@ describe('Rooms SDK Adapter', () => {
...
+        )
+        .subscribe((room) => {
           expect(room.title).toBe(updatedTitle);
           done();
-        }
-      });
+        });
     });
 
     test('support for multiple subscriptions', async (done) => {
```
```suggestion
    test('support for multiple subscriptions', (done) => {
```

PR #23, opened 2019-11-19
Define and Separate Unit and Integration coverage jobs

akoushke:


adamweeks:
Can you remove or update the commit message for the last commit?

package.json
```diff
@@ -13,8 +13,10 @@
     "release:dry-run": "semantic-release --dry-run",
     "start": "NODE_ENV=development nodemon scripts/start.js --exec babel-node",
     "test": "jest",
-    "test:integration": "jest \"^(?=.*integration.test.js).*\"",
+    "test:integration": "jest \"^(?=.*RoomsSDKAdapter.integration.test.js).*\"",
+    "test:integration:coverage": "npm run test:integration --coverage",
```
lalli-flores: Can't you just add the flag on the CI instead of creating another script for each?

.circleci/config.yml
```diff
@@ -43,7 +43,12 @@ jobs:
           command: npm run test:eslint
       - run:
           name: Unit Test Coverage
-          command: npm run test:coverage
+          command: npm run test:unit:coverage
+      - store_test_results:
+          path: test_results
+      - run:
```
lalli-flores: This should be its own job in the workflow, to run after and only, if linting + unit tests pass

akoushke: it will not get here if the unit test fails.

package.json
akoushke: yeah I will do that.

.circleci/config.yml
lalli-flores: It's easier to see exactly what failed if it's own step. If linting, unit tests and integration tests are separate, we can see right away from the PR which step failed, instead of having to go to CircleCi to find out what in "test" failed:
Screen Shot 2019-11-19 at 3 41 37 PM

adamweeks: If we’re following the widget and sdk pipeline style, the CI testing is one “workflow” with multiple jobs executed in order of: static analysis, unit tests, integration tests. I think it is worthwhile to keep the same flow in the component repos. Having them as separate jobs also makes it easier on circle ci to just re-run the failed job.

akoushke: where do you do the linting?

PR #24, opened 2019-11-22
Jest Puppeteer Installation and Configuration

akoushke:


src/jest-puppeteer.test.js
```diff
@@ -0,0 +1,15 @@
+import puppeteer from 'puppeteer';
```
akoushke: this file will be removed for the next PR.

.circleci/config.yml
```diff
@@ -1,10 +1,13 @@
 version: 2.1
+
 # reusable environment for all jobs
 executors:
   main-executor:
     working_directory: ~/sdk-component-adapter
     docker:
       - image: circleci/node:12.7.0
+        environment:
```
adamweeks: Should we utilize a browser image from circle? https://circleci.com/docs/2.0/browser-testing/#prerequisites

akoushke: how is that beneficial for our case?

adamweeks: I ask due to the title of the page linked: 

>This document describes common methods for running and debugging browser testing in your CircleCI config in the following sections:

Since we are adding browser testing, my question was should we use the tools provided.

akoushke: I see. Let's talk more about this.

src/testHelper.js
```diff
@@ -6,9 +6,9 @@ import {constructHydraId} from '@webex/common';
 // Used to get WEBEX_CLIENT_ID & WEBEX_CLIENT_SECRET for test user creation access
```
akoushke: do we still want to move to prod?

akoushke: @adamweeks do we still want to move to Prod?

scripts/puppeteer.server.js
```diff
@@ -0,0 +1,9 @@
+import * as http from 'http';
```
lalli-flores: If you can just import `createServer` since that's all we need
```suggestion
import {createServer} from 'http';
```

PR #31, opened 2019-12-12
Handle Audio/Video Controls

akoushke:
1 - mute/unmute for video/audio
2 - convert `then` to `async/await` 🙄 @lalli-flores 
3 - make control naming conventions more abstract

### DIFF
--- a/src/MeetingsSDKAdapter.js
+++ b/src/MeetingsSDKAdapter.js
@@ -66,7 +66,7 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
    *
    * @param {string} ID  ID to retrieve the SDK meeting object to add the local media to
    */
-  addLocalMedia(ID) {
+  async addLocalMedia(ID) {
     const sdkMeeting = this.fetchMeeting(ID);
 
     // default media setting
@@ -79,24 +79,18 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
       sendShare: false,
     };
 
-    sdkMeeting
-      .getMediaStreams(mediaSettings)
-      .then(([localStream, localShare]) => {
-        sdkMeeting
-          .addMedia({
-            localShare,
-            localStream,
-            mediaSettings,
-          })
-          .catch((error) => {
-            // eslint-disable-next-line no-console
-            console.error(`Unable to add local media to meeting "${ID}"`, error);
-          });
-      })
-      .catch((error) => {
-        // eslint-disable-next-line no-console
-        console.error(`Unable to get local media streams for meeting ${ID}`, error);
+    try {
+      const [localStream, localShare] = await sdkMeeting.getMediaStreams(mediaSettings);
+
+      await sdkMeeting.addMedia({
+        localShare,
+        localStream,
+        mediaSettings,
       });
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error(`Unable to add local media to meeting "${ID}"`, error);
+    }
   }
 
   /**
@@ -216,19 +210,18 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
    * @param {string} ID ID of the meeting to mute audio
    * @memberof MeetingsSDKAdapter
    */
-  muteAudioMeeting(ID) {
+  async muteAudioMeeting(ID) {
     const sdkMeeting = this.fetchMeeting(ID);
 
-    sdkMeeting
-      .muteAudio()
-      .then(() => {
-        // Due to SDK limitation around promises, we need to emit a custom event for audio mute action
-        sdkMeeting.emit(EVENT_MEDIA_LOCAL_MUTE, {control: MUTE_AUDIO_CONTROL});
-      })
-      .catch((error) => {
-        // eslint-disable-next-line no-console
-        console.error(`Unable to mute audio for meeting "${ID}"`, error);
-      });
+    try {
+      await sdkMeeting.muteAudio();
+
+      // Due to SDK limitation around promises, we need to emit a custom event for audio mute action
+      sdkMeeting.emit(EVENT_MEDIA_LOCAL_MUTE, {control: MUTE_AUDIO_CONTROL});
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error(`Unable to mute audio for meeting "${ID}"`, error);
+    }
   }
 
   /**
@@ -273,19 +266,18 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
    * @param {string} ID ID of the meeting to mute video
    * @memberof MeetingsSDKAdapter
    */
-  muteVideoMeeting(ID) {
+  async muteVideoMeeting(ID) {
     const sdkMeeting = this.fetchMeeting(ID);
 
-    sdkMeeting
-      .muteVideo()
-      .then(() => {
-        // Due to SDK limitation around promises, we need to emit a custom event for video mute action
-        sdkMeeting.emit(EVENT_MEDIA_LOCAL_MUTE, {control: MUTE_VIDEO_CONTROL});
-      })
-      .catch((error) => {
-        // eslint-disable-next-line no-console
-        console.error(`Unable to mute video for meeting "${ID}"`, error);
-      });
+    try {
+      await sdkMeeting.muteVideo();
+
+      // Due to SDK limitation around promises, we need to emit a custom event for video mute action
+      sdkMeeting.emit(EVENT_MEDIA_LOCAL_MUTE, {control: MUTE_VIDEO_CONTROL});
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error(`Unable to mute video for meeting "${ID}"`, error);
+    }
   }
 
   /**
--- a/src/MeetingsSDKAdapter.js
+++ b/src/MeetingsSDKAdapter.js
@@ -3,9 +3,9 @@ import {concat, from, fromEvent, Observable} from 'rxjs';
 import {filter, finalize, map, merge, publishReplay, refCount} from 'rxjs/operators';
 
 const EVENT_MEDIA_READY = 'media:ready';
-const EVENT_MEDIA_LOCAL_MUTE = 'adapter:media:local:mute';
+const EVENT_MEDIA_LOCAL_UPDATE = 'adapter:media:local:update';
 const JOIN_CONTROL = 'join-meeting';
-const MUTE_AUDIO_CONTROL = 'mute-audio';
+const AUDIO_CONTROL = 'audio';
 const MUTE_VIDEO_CONTROL = 'mute-video';
 const MEDIA_TYPE_LOCAL = 'local';
 const MEDIA_TYPE_REMOTE_AUDIO = 'remoteAudio';
@@ -31,10 +31,10 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
       display: this.joinControl.bind(this),
     };
 
-    this.meetingControls[MUTE_AUDIO_CONTROL] = {
-      ID: MUTE_AUDIO_CONTROL,
-      action: this.muteAudioMeeting.bind(this),
-      display: this.muteAudioControl.bind(this),
+    this.meetingControls[AUDIO_CONTROL] = {
+      ID: AUDIO_CONTROL,
+      action: this.handleLocalAudio.bind(this),
+      display: this.audioControl.bind(this),
     };
 
     this.meetingControls[MUTE_VIDEO_CONTROL] = {
@@ -210,17 +210,30 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
    * @param {string} ID ID of the meeting to mute audio
    * @memberof MeetingsSDKAdapter
    */
-  async muteAudioMeeting(ID) {
+  async handleLocalAudio(ID) {
     const sdkMeeting = this.fetchMeeting(ID);
 
     try {
-      await sdkMeeting.muteAudio();
+      let audioEnabled = this.meetings[ID].localAudio.getAudioTracks()[0].enabled;
 
-      // Due to SDK limitation around promises, we need to emit a custom event for audio mute action
-      sdkMeeting.emit(EVENT_MEDIA_LOCAL_MUTE, {control: MUTE_AUDIO_CONTROL});
+      if (audioEnabled) {
+        await sdkMeeting.muteAudio();
+      } else {
+        await sdkMeeting.unmuteAudio();
+      }
+
+      // re-assign the variable after the mute/unmute actions
+      audioEnabled = this.meetings[ID].localAudio.getAudioTracks()[0].enabled;
+
+      // Due to SDK limitation around local media updates,
+      // we need to emit a custom event for audio mute updates
+      sdkMeeting.emit(EVENT_MEDIA_LOCAL_UPDATE, {
+        control: AUDIO_CONTROL,
+        state: audioEnabled,
+      });
     } catch (error) {
       // eslint-disable-next-line no-console
-      console.error(`Unable to mute audio for meeting "${ID}"`, error);
+      console.error(`Unable to update local audio settings for meeting "${ID}"`, error);
     }
   }
 
@@ -231,32 +244,39 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
    * @returns {Observable.<MeetingControlDisplay>}
    * @memberof MeetingJSONAdapter
    */
-  muteAudioControl(ID) {
+  audioControl(ID) {
     const sdkMeeting = this.fetchMeeting(ID);
     const muted = {
-      ID: MUTE_AUDIO_CONTROL,
-      icon: 'microphone',
-      tooltip: 'Mute',
+      ID: AUDIO_CONTROL,
+      icon: 'microphone-muted',
+      tooltip: 'Unmute',
       state: MeetingControlState.ACTIVE,
+      text: null,
+    };
+    const unmuted = {
+      ID: AUDIO_CONTROL,
+      icon: 'microphone-muted',
+      tooltip: 'Mute',
+      state: MeetingControlState.INACTIVE,
       text: null,
     };
 
     const getDisplayData$ = Observable.create((observer) => {
       if (sdkMeeting) {
-        observer.next(muted);
+        observer.next(unmuted);
       } else {
-        observer.error(new Error(`Could not find meeting with ID "${ID}" to mute audio on`));
+        observer.error(new Error(`Could not find meeting with ID "${ID}" to add audio control`));
       }
 
       observer.complete();
     });
 
-    const muteEvent$ = fromEvent(sdkMeeting, EVENT_MEDIA_LOCAL_MUTE).pipe(
-      filter((event) => event.control === MUTE_AUDIO_CONTROL),
-      map(() => ({...muted, state: MeetingControlState.INACTIVE}))
+    const localMediaUpdateEvent$ = fromEvent(sdkMeeting, EVENT_MEDIA_LOCAL_UPDATE).pipe(
+      filter((event) => event.control === AUDIO_CONTROL),
+      map(({state}) => (state ? unmuted : muted))
     );
 
-    return concat(getDisplayData$, muteEvent$);
+    return concat(getDisplayData$, localMediaUpdateEvent$);
   }
 
   /**
@@ -273,7 +293,7 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
       await sdkMeeting.muteVideo();
 
       // Due to SDK limitation around promises, we need to emit a custom event for video mute action
-      sdkMeeting.emit(EVENT_MEDIA_LOCAL_MUTE, {control: MUTE_VIDEO_CONTROL});
+      sdkMeeting.emit(EVENT_MEDIA_LOCAL_UPDATE, {control: MUTE_VIDEO_CONTROL});
     } catch (error) {
       // eslint-disable-next-line no-console
       console.error(`Unable to mute video for meeting "${ID}"`, error);
@@ -307,7 +327,7 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
       observer.complete();
     });
 
-    const muteEvent$ = fromEvent(sdkMeeting, EVENT_MEDIA_LOCAL_MUTE).pipe(
+    const muteEvent$ = fromEvent(sdkMeeting, EVENT_MEDIA_LOCAL_UPDATE).pipe(
       filter((event) => event.control === MUTE_VIDEO_CONTROL),
       map(() => ({...muted, state: MeetingControlState.INACTIVE}))
     );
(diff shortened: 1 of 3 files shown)


src/MeetingsSDKAdapter.js
```diff
@@ -355,8 +380,17 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
 
+      // Listen to unmute event to return the meeting object
+      const meetingWithLocalUnmuteEvents$ = fromEvent(sdkMeeting, EVENT_MEDIA_LOCAL_UNMUTE).pipe(
+        map(() => this.meetings[ID])
+      );
+
       // Merge all event observables to update the existing meeting object simultaneously
-      const meetingsWithEvents$ = merge(meetingWithReadyEvent$, meetingWithLocalMuteEvents$);
+      const meetingsWithEvents$ = merge(
+        meetingWithReadyEvent$,
```
adamweeks: Could you rename this to `meetingWithMediaReadyEvent$`? I know it isn’t a part of this PR but it keeps confusing me

src/MeetingsSDKAdapter.js
```diff
@@ -79,24 +80,23 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
-          .catch((error) => {
-            // eslint-disable-next-line no-console
-            console.error(`Unable to add local media to meeting "${ID}"`, error);
-          });
-      })
-      .catch((error) => {
+    try {
+      const [localStream, localShare] = await sdkMeeting.getMediaStreams(mediaSettings);
+
+      try {
```
adamweeks: Should we move this outside the `getMediaStreams` try block? I don't think it has to be nested does it?

src/MeetingsSDKAdapter.js
```diff
@@ -259,8 +273,8 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
     });
 
     const muteEvent$ = fromEvent(sdkMeeting, EVENT_MEDIA_LOCAL_MUTE).pipe(
-      filter((event) => event.control === MUTE_AUDIO_CONTROL),
-      map(() => ({...muted, state: MeetingControlState.INACTIVE}))
+      filter((event) => event.control === AUDIO_CONTROL),
+      map(({state}) => (state ? muted : unmuted))
```
adamweeks: Where does this state come from?

src/MeetingsSDKAdapter.js
```diff
@@ -273,19 +287,26 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
    * @param {string} ID ID of the meeting to mute video
    * @memberof MeetingsSDKAdapter
    */
-  muteVideoMeeting(ID) {
+  async handleLocalVideo(ID) {
     const sdkMeeting = this.fetchMeeting(ID);
+    const state = this.meetings[ID].localVideo.getVideoTracks()[0].enabled;
```
adamweeks: This could use some error handling around checking if `localVideo` is an object and has the `getVideoTracks` function.

src/MeetingsSDKAdapter.js
akoushke: I had the blocks, to catch each errors respectively. base on my knowledge, each `catch` needs a `try`.

src/MeetingsSDKAdapter.js
akoushke: this object should exist when we are at this point.

src/MeetingsSDKAdapter.js
```diff
@@ -344,7 +372,7 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
       });
 
       // Listen to attach mediaStream source objects to the existing meeting
-      const meetingWithReadyEvent$ = fromEvent(sdkMeeting, EVENT_MEDIA_READY).pipe(
+      const meetingWithMediaReadyEvent$ = fromEvent(sdkMeeting, EVENT_MEDIA_READY).pipe(
```
adamweeks: 👍 thanks

src/MeetingsSDKAdapter.js
akoushke: so this comes from line 234, where we are emitting the event with the given object which contains `state` prop

src/MeetingsSDKAdapter.js
lalli-flores: I don't think it's necessary to have the nested `try`s to distinguish between errors. You could just have one `catch` for either error and just log out `Unable to add local media to meeting "${ID}"`` which is what this method is trying to do.

src/MeetingsSDKAdapter.js
akoushke: basically we have two listeners, one in `getMeeting` and one in `audioControl` and they both listen to the changes emitted from `hanldeLocalAudio` method

src/MeetingsSDKAdapter.test.js
```diff
@@ -157,11 +157,44 @@ describe('Meetings SDK Adapter', () => {
...
+        ...meeting,
+        localAudio: {
+          getAudioTracks: jest.fn(() => [{enabled: true}]),
+        },
+      };
+    });
+
+    test('mutes audio if the the audio track is enabled', () => {
+      meetingSDKAdapter.handleLocalAudio(meetingID);
+      expect(mockSDKMeeting.muteAudio).toBeCalled();
```
adamweeks: Since `toBeCalled` is just an alias for `toHaveBeenCalled` and we're using `toHaveBeenCalled` elsewhere in this repo, can we standardize on using `toHaveBeenCalled`?

adamweeks:
Thanks for adding those tests, our coverage went from 69% to  86%!

src/MeetingsSDKAdapter.js
```diff
@@ -216,19 +216,26 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
-        // eslint-disable-next-line no-console
-        console.error(`Unable to mute audio for meeting "${ID}"`, error);
-      });
+    try {
+      if (state) {
+        await sdkMeeting.muteAudio();
+        eventName = EVENT_MEDIA_LOCAL_MUTE;
+      } else {
+        await sdkMeeting.unmuteAudio();
+        eventName = EVENT_MEDIA_LOCAL_UNMUTE;
```
lalli-flores: Why do you need to have two separate events? You can distinguish from the data passed in what you need to do without listening/emitting for two separate events

src/MeetingsSDKAdapter.test.js
akoushke: sure

src/MeetingsSDKAdapter.js
```diff
@@ -238,15 +245,22 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
    * @returns {Observable.<MeetingControlDisplay>}
    * @memberof MeetingJSONAdapter
    */
-  muteAudioControl(ID) {
+  audioControl(ID) {
     const sdkMeeting = this.fetchMeeting(ID);
     const muted = {
```
lalli-flores: The states seemed to be flipped. See [JSON implementation](https://github.com/webex/components/blob/master/src/adapters/MeetingsJSONAdapter.js) for reference

src/MeetingsSDKAdapter.js
adamweeks: My suggestion is to do the two `try`s separately:

```js
try {
 await getMediaStreams
} catch (error) {
}

try {
 await addMedia
} catch (error) {
```

src/MeetingsSDKAdapter.js
lalli-flores: Can you name this variable `enabled` or `videoEnabled`? `if/else` is less readable with the word `state`

src/MeetingsSDKAdapter.js
```diff
@@ -79,24 +80,23 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
-            // eslint-disable-next-line no-console
-            console.error(`Unable to add local media to meeting "${ID}"`, error);
-          });
-      })
-      .catch((error) => {
+    try {
+      const [localStream, localShare] = await sdkMeeting.getMediaStreams(mediaSettings);
+
+      try {
+        sdkMeeting.addMedia({
```
adamweeks: I don't think you will catch a promise thrown error in a try/catch block without adding `async sdkMeeting.addMedia`.

src/MeetingsSDKAdapter.js
```diff
@@ -295,15 +316,22 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
     const muted = {
-      ID: MUTE_VIDEO_CONTROL,
+      ID: VIDEO_CONTROL,
       icon: 'camera',
       tooltip: 'Mute',
       state: MeetingControlState.ACTIVE,
       text: null,
     };
+    const unmuted = {
+      ID: VIDEO_CONTROL,
```
lalli-flores: The values for this display seemed to be flipped

src/MeetingsSDKAdapter.js
adamweeks: Ok, so it is from our constructed sdk event. I didn't remember seeing the `state` property on the native sdk events.

src/MeetingsSDKAdapter.js
akoushke: I ended up using one try block. good catch.

src/MeetingsSDKAdapter.js
adamweeks: Just because it should have the proper function, doesn't mean it will. This could be a hard crash if localvideo is undefined.

src/MeetingsSDKAdapter.js
```diff
@@ -216,19 +210,24 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
-        console.error(`Unable to mute audio for meeting "${ID}"`, error);
-      });
+    try {
+      if (audioEnabled) {
+        await sdkMeeting.muteAudio();
+      } else {
+        await sdkMeeting.unmuteAudio();
+      }
+
+      // Due to SDK limitation around promises, we need to emit a custom event for audio mute action
```
lalli-flores:
```suggestion
      // Due to SDK limitation around local media updates,
      // we need to emit a custom event for audio mute updates
```

src/MeetingsSDKAdapter.js
```diff
@@ -216,19 +210,24 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
+      sdkMeeting.emit(eventName, {control: AUDIO_CONTROL, state: !audioEnabled});
```
lalli-flores: Where is `eventName` set?

src/MeetingsSDKAdapter.js
```diff
@@ -216,19 +210,24 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error(`Unable to handle local audio for meeting "${ID}"`, error);
```
lalli-flores:
```suggestion
      console.error(`Unable to update local audio settings for meeting "${ID}"`, error);
```

src/MeetingsSDKAdapter.js
```diff
@@ -258,9 +264,9 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
 
-    const muteEvent$ = fromEvent(sdkMeeting, EVENT_MEDIA_LOCAL_MUTE).pipe(
-      filter((event) => event.control === MUTE_AUDIO_CONTROL),
-      map(() => ({...muted, state: MeetingControlState.INACTIVE}))
+    const muteEvent$ = fromEvent(sdkMeeting, EVENT_MEDIA_LOCAL_UPDATE).pipe(
+      filter((event) => event.control === AUDIO_CONTROL),
+      map(({state}) => (state ? muted : unmuted))
     );
 
     return concat(getDisplayData$, muteEvent$);
```
lalli-flores: I can't comment on the exact line, but make sure that `getDisplayData$` send `unmuted` display as default. When we first load the control, it should not be activated

src/MeetingsSDKAdapter.js
lalli-flores: Would it be better to re-assign `audioEnabled` after muting/unmuting (in the `if/else`?
This way we send the true current state of the audio track?

src/MeetingsSDKAdapter.js
```diff
@@ -273,19 +279,24 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
-      });
+    try {
+      if (videoEnabled) {
+        await sdkMeeting.muteVideo();
+      } else {
+        await sdkMeeting.unmuteVideo();
+      }
+
+      // Due to SDK limitation around promises, we need to emit a custom event for video mute action
+      sdkMeeting.emit(eventName, {control: VIDEO_CONTROL, state: !videoEnabled});
```
lalli-flores: Similar comments as in the audio handling method:
* Where is `eventName` set?
* should you re-assign `videoEnabled` after muting/unmuting so we send the current state of the SDK/server track?

src/MeetingsSDKAdapter.js
```diff
@@ -273,19 +279,24 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error(`Unable to handle local video for meeting "${ID}"`, error);
```
lalli-flores:
```suggestion
      console.error(`Unable to update local video settings for meeting "${ID}"`, error);
```

src/MeetingsSDKAdapter.js
```diff
@@ -295,13 +306,20 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
+      ID: VIDEO_CONTROL,
+      icon: 'camera-muted',
+      tooltip: 'Unmute',
+      state: MeetingControlState.ACTIVE,
+      text: null,
+    };
+    const unmuted = {
+      ID: VIDEO_CONTROL,
       icon: 'camera',
       tooltip: 'Mute',
```
lalli-flores: From UX, the tooltip should say `Stop video`

src/MeetingsSDKAdapter.js
adamweeks: That works. If we need to be more specific as to which function failed, we would need to do two try blocks.

src/MeetingsSDKAdapter.js
akoushke: I started the discussion in the space.

src/MeetingsSDKAdapter.js
```diff
@@ -295,13 +306,20 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
-      ID: MUTE_VIDEO_CONTROL,
+      ID: VIDEO_CONTROL,
+      icon: 'camera-muted',
+      tooltip: 'Unmute',
+      state: MeetingControlState.ACTIVE,
+      text: null,
+    };
+    const unmuted = {
+      ID: VIDEO_CONTROL,
       icon: 'camera',
```
lalli-flores: I just re-checked the UX design, for both, mute and unmute states, the icon should be `camera-muted`

src/MeetingsSDKAdapter.js
```diff
@@ -295,13 +306,20 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
    * @memberof MeetingJSONAdapter
    */
-  muteVideoControl(ID) {
+  videoControl(ID) {
     const sdkMeeting = this.fetchMeeting(ID);
     const muted = {
-      ID: MUTE_VIDEO_CONTROL,
+      ID: VIDEO_CONTROL,
+      icon: 'camera-muted',
+      tooltip: 'Unmute',
```
lalli-flores: This should say `Start video`

src/MeetingsSDKAdapter.js
```diff
@@ -238,13 +237,20 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
-      ID: MUTE_AUDIO_CONTROL,
+      ID: AUDIO_CONTROL,
+      icon: 'microphone-muted',
+      tooltip: 'Unmute',
+      state: MeetingControlState.ACTIVE,
+      text: null,
+    };
+    const unmuted = {
+      ID: AUDIO_CONTROL,
       icon: 'microphone',
```
lalli-flores: I just re-checked UX and for unmute state the icon should also be `microphone-muted`

src/MeetingsSDKAdapter.js
```diff
@@ -315,12 +333,12 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
       observer.complete();
     });
 
-    const muteEvent$ = fromEvent(sdkMeeting, EVENT_MEDIA_LOCAL_MUTE).pipe(
-      filter((event) => event.control === MUTE_VIDEO_CONTROL),
-      map(() => ({...muted, state: MeetingControlState.INACTIVE}))
+    const handleEvent$ = fromEvent(sdkMeeting, EVENT_MEDIA_LOCAL_UPDATE).pipe(
```
lalli-flores: Why isn't this called `muteEvent$` like in the other methods?

src/MeetingsSDKAdapter.js
```diff
@@ -315,12 +333,12 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
+      filter((event) => event.control === VIDEO_CONTROL),
+      map(({state}) => (state ? muted : unmuted))
     );
 
-    return concat(getDisplayData$, muteEvent$);
+    return concat(getDisplayData$, handleEvent$);
```
lalli-flores: Make sure that the default display is unmuted, as the control will start out inactive

src/MeetingsSDKAdapter.js
```diff
@@ -344,19 +362,19 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
+      const meetingWithMediaReadyEvent$ = fromEvent(sdkMeeting, EVENT_MEDIA_READY).pipe(
         filter((event) => MEDIA_EVENT_TYPES.includes(event.type)),
         map((event) => this.attachMedia(ID, event)),
         map(() => this.meetings[ID])
       );
 
-      // Listen to mute event to return the meeting object
-      const meetingWithLocalMuteEvents$ = fromEvent(sdkMeeting, EVENT_MEDIA_LOCAL_MUTE).pipe(
+      // Listen to update event to return the meeting object
+      const meetingWithLocalUpdateEvents$ = fromEvent(sdkMeeting, EVENT_MEDIA_LOCAL_UPDATE).pipe(
```
lalli-flores: I thought you said updating the meeting object wasn't necessary? Since the meeting object reference that the component will have would also get updated

akoushke: we still need to push the data so the UI knows there had been a change.

src/MeetingsSDKAdapter.test.js
```diff
@@ -184,6 +227,39 @@ describe('Meetings SDK Adapter', () => {
...
+        },
+      };
+    });
+
+    test('mutes video if the the video track is enabled', () => {
+      meetingSDKAdapter.handleLocalVideo(meetingID);
+      expect(mockSDKMeeting.muteVideo).toHaveBeenCalled();
+    });
+
+    test('unmutes video if the the video track is enabled', () => {
```
adamweeks:
```suggestion
    test('unmutes video if the the video track is disabled', () => {
```

src/MeetingsSDKAdapter.js
akoushke: why does that matter?

src/MeetingsSDKAdapter.js
akoushke: will take care of this in upcoming PR

src/MeetingsSDKAdapter.js
```diff
@@ -238,32 +238,39 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
+      state: MeetingControlState.INACTIVE,
+      text: null,
+    };
 
     const getDisplayData$ = Observable.create((observer) => {
       if (sdkMeeting) {
-        observer.next(muted);
+        observer.next(unmuted);
       } else {
         observer.error(new Error(`Could not find meeting with ID "${ID}" to mute audio on`));
```
lalli-flores:
```suggestion
        observer.error(new Error(`Could not find meeting with ID "${ID}" to add audio control`));
```

src/MeetingsSDKAdapter.js
```diff
@@ -238,32 +238,39 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
 
       observer.complete();
     });
 
-    const muteEvent$ = fromEvent(sdkMeeting, EVENT_MEDIA_LOCAL_MUTE).pipe(
-      filter((event) => event.control === MUTE_AUDIO_CONTROL),
-      map(() => ({...muted, state: MeetingControlState.INACTIVE}))
+    const localMediaUpdateEvent$ = fromEvent(sdkMeeting, EVENT_MEDIA_LOCAL_UPDATE).pipe(
+      filter((event) => event.control === AUDIO_CONTROL),
+      map(({state}) => (state ? muted : unmuted))
```
lalli-flores: Shouldn't be the other way around?
if the `state` is `enabled` we want to send `unmuted` because audio is enabled, and `disabled` means muted, no?

src/MeetingsSDKAdapter.js
```diff
@@ -295,32 +308,39 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
+      state: MeetingControlState.INACTIVE,
+      text: null,
+    };
 
     const getDisplayData$ = Observable.create((observer) => {
       if (sdkMeeting) {
-        observer.next(muted);
+        observer.next(unmuted);
       } else {
         observer.error(new Error(`Could not find meeting with ID "${ID}" to mute video on`));
```
lalli-flores:
```suggestion
        observer.error(new Error(`Could not find meeting with ID "${ID}" to add video control`));
```

src/MeetingsSDKAdapter.test.js
```diff
@@ -147,24 +157,57 @@ describe('Meetings SDK Adapter', () => {
...
-          expect(error.message).toBe('Could not find meeting with ID "meetingID" to mute audio on');
+          expect(error.message).toBe('Could not find meeting with ID "meetingID" to add audio control');
           done();
         }
       );
     });
   });
 
-  describe('muteVideoControl()', () => {
+  describe('handleLocalAudio()', () => {
```
lalli-flores: Can you add a test that check that the event gets emitted?

src/MeetingsSDKAdapter.test.js
```diff
@@ -174,16 +217,49 @@ describe('Meetings SDK Adapter', () => {
...
         (error) => {
-          expect(error.message).toBe('Could not find meeting with ID "meetingID" to mute video on');
+          expect(error.message).toBe('Could not find meeting with ID "meetingID" to add video control');
           done();
         }
       );
     });
   });
 
+  describe('handleLocalVideo()', () => {
```
lalli-flores: Can you add a test that check that the event gets emitted?

src/MeetingsSDKAdapter.test.js
```diff
@@ -34,6 +34,16 @@ describe('Meetings SDK Adapter', () => {
     target = null;
   });
 
+  describe('addLocalMedia()', () => {
+    test('throws errors if the local media is not added to the meeting successfully', async () => {
+      mockSDKMeeting.getMediaStreams = jest.fn(() => Promise.reject());
+      global.console.error = jest.fn();
+      await meetingSDKAdapter.addLocalMedia(meetingID);
+
+      expect(global.console.error).toHaveBeenCalledWith('Unable to add local media to meeting "meetingID"', undefined);
```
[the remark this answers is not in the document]
akoushke: the error that will be sent

PR #34, opened 2019-12-14
Implement getMe

akoushke:


src/PeopleSDKAdapter.js
```diff
@@ -53,6 +53,17 @@ export default class PeopleSDKAdapter extends PeopleAdapter {
     };
   }
 
+  /**
+   * Returns an observable that emits me data of.
```
lalli-flores:
```suggestion
   * Returns an observable that emits person data of the access token bearer.
```

src/PeopleSDKAdapter.js
```diff
@@ -53,6 +53,17 @@ export default class PeopleSDKAdapter extends PeopleAdapter {
...
+   *
+   * @public
+   * @returns {Observable.<Person>}
+   * @memberof PeopleSDKAdapter
+   */
+  getMe() {
+    return from(this.datasource.people.get('me')).pipe(flatMap(({id}) => this.getPerson(id)));
```
lalli-flores: Why bother calling `people.get` at all? Can't you just return `this.getPerson('me')`?

src/PeopleSDKAdapter.test.js
```diff
@@ -12,6 +12,24 @@ describe('People SDK Adapter', () => {
     personID = 'personID';
   });
 
+  describe('getMe()', () => {
+    test('returns a person in a proper shape', (done) => {
+      peopleSDKAdapter.getMe().subscribe((person) => {
+        expect(person).toEqual({
```
lalli-flores:
```suggestion
        expect(person).toMatchObject({
```

src/PeopleSDKAdapter.js
akoushke: to listen to all status changes. and return the mapped person Model.

akoushke: just re-read your comment. we can't do that, because `deconstructHydraId` method invoked inline 78 inside `getPerson` method definition from the common library expects an actual ID

PR #36, opened 2019-12-17
Jest Meeting Dev Server

akoushke:


scripts/puppeteer.server.js
```diff
@@ -1,7 +1,81 @@
...
+    console.error('Unable to dial:', error);
+  }
+});
+
+document.getElementById('actions').addEventListener('click', async (event) => {
+  event.preventDefault();
+
+  try {
+    switch (event.target.name) {
+      case 'mute-audio':
```
lalli-flores: Does it makes sense to separate `mute-audio` and `unmute-audio`? Since both call the same function, you can't guarantee that each button does what they are labeled to do. For instance, if you click mute and then you click it again,  even though it says mute, it will do an "unmute"

package.json
```diff
@@ -39,7 +39,6 @@
     "@babel/cli": "^7.6.4",
     "@babel/core": "^7.6.4",
     "@babel/node": "^7.6.3",
-    "@babel/plugin-transform-regenerator": "^7.4.5",
```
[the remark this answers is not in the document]
akoushke: it's not needed. Also `Parcel` have this configured already. Not sure why we installed it at the first place.

src/MeetingsSDKAdapter.js
```diff
@@ -113,9 +113,11 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
         };
         break;
       case MEDIA_TYPE_REMOTE_AUDIO:
+        console.log('here');
         this.meetings[ID] = {...meeting, remoteAudio: stream};
         break;
       case MEDIA_TYPE_REMOTE_VIDEO:
+        console.log('here');
```
InteractiveTimmy: Not sure if you'd want these `console.log()` statements.

PR #37, opened 2019-12-17
Improve getMe()

akoushke:
getMe() should only fetches and emits the data once!

### DIFF
--- a/src/PeopleSDKAdapter.integration.test.js
+++ b/src/PeopleSDKAdapter.integration.test.js
@@ -28,6 +28,19 @@ describe('People SDK Adapter', () => {
     }
   });
 
+  describe('getMe()', () => {
+    test('returns person data of the access token bearer in a proper shape', (done) => {
+      subscription = webexSDKAdapter.peopleAdapter.getMe().subscribe((person) => {
+        expect(person).toMatchObject({
+          ID: userID,
+          emails: [user.emailAddress],
+          displayName: user.displayName,
+        });
+        done();
+      });
+    });
+  });
+
   describe('getPerson() returns', () => {
     beforeEach(() => {
       getPerson$ = webexSDKAdapter.peopleAdapter.getPerson(userID);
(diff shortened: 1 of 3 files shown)


src/PeopleSDKAdapter.js
```diff
@@ -54,15 +54,21 @@ export default class PeopleSDKAdapter extends PeopleAdapter {
   }
 
   /**
-   * Returns an observable that emits person data of the access token bearer.
+   * Returns an observable that emits only once of person data of the access token bearer.
```
lalli-flores: I would prefer to call out that the observable completes in a separate line.
```suggestion
   * Returns an observable that emits person data of the access token bearer.
   * The observable will emit once and then complete.
```

src/PeopleSDKAdapter.js
```diff
@@ -54,15 +54,21 @@ export default class PeopleSDKAdapter extends PeopleAdapter {
...
    * @returns {Observable.<Person>}
    * @memberof PeopleSDKAdapter
    */
   getMe() {
-    // person ID must be retrieved in order to invoke `getPerson` method properly.
-    return from(this.datasource.people.get('me')).pipe(flatMap(({id}) => this.getPerson(id)));
+    return from(this.fetchPerson('me')).pipe(
+      flatMap((person) =>
+        from(this.datasource.internal.presence.get([person.id])).pipe(
+          map((status) => ({...person, status: this.getStatus(status.status)}))
```
lalli-flores: If you call it `presence` it would read better
```suggestion
          map((presence) => ({...person, status: this.getStatus(presence.status)}))
```

src/PeopleSDKAdapter.integration.test.js
```diff
@@ -28,6 +28,19 @@ describe('People SDK Adapter', () => {
     }
   });
 
+  describe('getMe()', () => {
+    test('a person data of the access token bearer in a proper shape', (done) => {
```
lalli-flores:
```suggestion
    test('returns person data of the access token bearer in a proper shape', (done) => {
```

lalli-flores:
Can you squash both commits as one?

PR #38, opened 2019-12-17
fix(package): set webex packages to 1.80.43

adamweeks:


adamweeks:
@akoushke dismissed your review because it was using the latest version of the sdk, but that appears to be problematic, so setting the version to the current version. 

@lalli-flores we should create a JIRA for updating to the very latest version of the sdk, but it will be more work due to some service catalog issues.

akoushke:
so the aphelia service works here?

adamweeks:
@akoushke 

> so the aphelia service works here?

yes, works fine after sticking the version.

PR #39, opened 2019-12-18
docs(Readme): update code samples for react

adamweeks:


README.md
```diff
@@ -34,36 +35,160 @@ npm install --save @webex/components @webex/sdk-component-adapter
...
+These connections are handled by the SDK Adapter, but require the `connect` function to be manually called.
+
+```js
+const webex = new Webex({
+  credentials: `<YOUR WEBEX ACCESS_TOKEN>`,
+});
+
+const adapter = new WebexSDKAdapter(webex);
+
+adapter.connect();
```
akoushke: I think we should let the user know that they need to wait for this promise to be resolved. by using `async/await`. Otherwise the devices will not be registered properly.

README.md
```diff
@@ -34,36 +35,160 @@ npm install --save @webex/components @webex/sdk-component-adapter
...
+```
+
+#### Disconnect
+
+When the adapter is no longer going to be used, the `disconnect` function will do the necessary tear-down of all the connections created in the `connect` function:
+
+```js
+adapter.disconnect();
```
akoushke: same here ^

README.md
```diff
@@ -34,36 +35,160 @@ npm install --save @webex/components @webex/sdk-component-adapter
...
+  const [isReady, setReady] = useState(false);
+
+  useEffect(() => {
+    // This is the suggested way to do async hooks.
+    // Ref: https://github.com/facebook/react/issues/14326
+    async function doConnect() {
+      const adapter = new WebexSDKAdapter(webex);
+      // Wait for the adapter to connect before rendering.
+      await adapter.connect();
+      setReady(true);
```
akoushke: maybe use `isConnected`? something more descriptive?

README.md
```diff
@@ -34,36 +35,160 @@ npm install --save @webex/components @webex/sdk-component-adapter
...
+const webex = new Webex({
+  credentials,
+});
+
+export default class App extends Component {
+
+  constructor(props) {
+    super(props);
+
+    this.state = {isReady: false};
```
akoushke: maybe use some more descriptive flag name? `isConnected`?

README.md
```diff
@@ -21,6 +21,7 @@
 
 - [Install](#install)
 - [Usage](#usage)
+  - [Setup](#setup)
```
akoushke: are you planning to add the the other titles that you defined in the midsection here as well?

README.md
```diff
@@ -34,36 +35,160 @@ npm install --save @webex/components @webex/sdk-component-adapter
...
 
-const adapter = new WebexSDKAdapter(webex);
+function App() {
+  const [isReady, setReady] = useState(false);
+
+  useEffect(() => {
+    // This is the suggested way to do async hooks.
+    // Ref: https://github.com/facebook/react/issues/14326
+    async function doConnect() {
+      const adapter = new WebexSDKAdapter(webex);
```
mwegman: You need to move the declaration of `adapter` out of this function so the cleanup call below can use it.

README.md
```diff
@@ -34,36 +40,167 @@ npm install --save @webex/components @webex/sdk-component-adapter
...
+      await adapter.connect();
+      setAdapterConnected(true);
+    }
+
+    doConnect();
+
+    // On teardown, disconnect the adapter
+    return async () => {
+      await adapter.disconnect();
+    }
```
mwegman: This function doesn't need to be async/await. Nothing is happening after the disconnect call.

README.md
```diff
@@ -34,36 +40,167 @@ npm install --save @webex/components @webex/sdk-component-adapter
...
+      // Wait for the adapter to connect before rendering.
+      await adapter.connect();
+      setAdapterConnected(true);
+    }
+
+    doConnect();
+
+    // On teardown, disconnect the adapter
+    return async () => {
+      await adapter.disconnect();
```
lalli-flores: Don't forget to `setAdapterConnected` to `false`

adamweeks: If the component is being removed, what purpose does updating the state do?

README.md
```diff
@@ -34,36 +40,167 @@ npm install --save @webex/components @webex/sdk-component-adapter
...
+    super(props);
+
+    this.state = {
+      adapter: null,
+      adapterConnected: false
+    };
+  }
+
+  async componentDidMount() {
+    const adapter = new WebexSDKAdapter(webex);
```
akoushke: I think this should be a class property and defined the constructor.

```js
const webex = new Webex({
  credentials,
});
this.adapter = new WebexSDK(webex);
```

PR #44, opened 2020-01-16
fix(MeetingSdkAdapter): emit stopped events for remote media

akoushke:


src/MeetingsSDKAdapter.js
```diff
@@ -229,13 +229,19 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
-        console.error(`Unable to leave from the meeting "${ID}"`, error);
-      });
+  async leaveMeeting(ID) {
+    try {
+      const sdkMeeting = this.fetchMeeting(ID);
+
+      await sdkMeeting.leave();
+
+      // Due to SDK limitations, We need to emit a media stopped event for remote media types
+      sdkMeeting.emit(EVENT_MEDIA_STOPPED, {type: MEDIA_TYPE_REMOTE_AUDIO});
```
adamweeks: Can we use the same pattern we did with `EVENT_MEDIA_LOCAL_UPDATE` to not use the same events the sdk is **supposed** to emit? That way if it gets fixed we're not getting duplicate events.

akoushke: I find that pretty redundant, since we have the functionality implemented for this already. we are handling adding local media the same way by triggering the the same event after the local streams have been added to the meeting object.

akoushke: Also custom event was created since there was no other events supporting this feature. If we come up with a different custom event for, We will need to create another listener to do the same thing that we are doing for local stopped event which I find redundant.

adamweeks: I guess that’s fine. Even if we upgrade to a version that actually has media stopped events, it shouldn’t cause any hard to have duplicate events since it will just set null to null anyway.

PR #45, opened 2020-01-24
feat(MeetingsSdkAdapter): add the capability to join the meeting muted

akoushke:


src/MeetingsSDKAdapter.js
```diff
@@ -210,12 +212,21 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
     try {
       const sdkMeeting = this.fetchMeeting(ID);
       const {localAudio, localVideo} = this.meetings[ID];
-      const localStream = new MediaStream([localAudio.getAudioTracks()[0], localVideo.getVideoTracks()[0]]);
+      const [localStream] = await sdkMeeting.getMediaStreams(DEFAULT_MEDIA_SETTINGS);
 
       await sdkMeeting.join();
 
       // SDK requires to join the meeting before adding the local stream media to the meeting
       await sdkMeeting.addMedia({localStream, DEFAULT_MEDIA_SETTINGS});
```
adamweeks: If both local audio and video are muted at this point, we could skip calling `addMedia`, that way we wouldn't have to immediately mute after joining. But, we would need some sort of way to actually add the media once the user unmutes. It would require utilizing the sdk meeting's state machine to determine if there is actually local media.

src/MeetingsSDKAdapter.js
```diff
@@ -295,16 +306,25 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
     const sdkMeeting = this.fetchMeeting(ID);
 
     try {
+      const isInSession = this.meetings[ID].remoteAudio !== null;
```
adamweeks: You can more accurately determine if the meeting is joined by inspecting the `meeting.meetingFiniteStateMachine.state` property. The constant to check would be `MEETING_STATE_MACHINE.STATES.JOINED`

src/MeetingsSDKAdapter.js
```diff
@@ -376,16 +396,25 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
     const sdkMeeting = this.fetchMeeting(ID);
 
     try {
+      const isInSession = this.meetings[ID].remoteVideo !== null;
```
adamweeks: It might be worthwhile to make this a helper function like `isMeetingJoined(ID)`

src/MeetingsSDKAdapter.js
adamweeks: Probably a future improvement, not required in this PR.

src/MeetingsSDKAdapter.test.js
```diff
@@ -59,16 +61,57 @@ describe('Meetings SDK Adapter', () => {
       };
 
       global.MediaStream = jest.fn(() => mockMediaStreamInstance);
+    });
+
+    test('keep `localAudio` empty, if the event type is `local` and the audio stream is empty', () => {
```
adamweeks: Would like to see these added tests grouped in a describe with `describe('when the event type is local and the audio stream is empty', () => {`

src/MeetingsSDKAdapter.js
akoushke: that was my initial plan. But sdk `AddMedia` method does not behave properly when I unmute on a non-existent stream. that's why I just decided to mute the streams instead.

src/MeetingsSDKAdapter.test.js
akoushke: not sure if we need to group this on a different `describe`. generally we are grouping the test cases by methods. and all these test cases fall under `attacheMedia` method/describe.

src/MeetingsSDKAdapter.js
akoushke: is this publicly supported? I might have missed it in the docs.

src/MeetingsSDKAdapter.js
akoushke: this should be implemented once we have a legit property - AKA state/meetingState

PR #46, opened 2020-01-24
fix(MeetingSdkAdapter): mute controls are now synced up properly

akoushke:


lalli-flores:
Closed PR because the original values were correct

PR #48, opened 2020-01-25
chore(WebApp): remove local audio

akoushke:


scripts/index.html
```diff
@@ -34,14 +34,14 @@ <h1 id="page-title">
         <div style="display: flex; justify-content: space-between;">
           <div>
             <h2>Local Stream:</h2>
-            <audio id="local-audio" autoplay playsinline></audio>
+            <!-- <audio id="local-audio" autoplay playsinline></audio> -->
```
lalli-flores: Why don't you remove it, instead of commenting it out?
Or, we could mute it maybe? so we can still do testing for local audio

scripts/index.html
```diff
@@ -34,14 +34,14 @@ <h1 id="page-title">
...
             <video style="width:25rem" id="local-video" autoplay playsinline></video>
           </div>
           <h2 id="meeting-title"></h2>
           <div>
             <h2>Remote Stream:</h2>
             <audio id="remote-audio" autoplay playsinline></audio>
-            <video style="width:25rem" id="remote-video" autoplay playsinline></video>
+            <video style="width:25rem" id="remote-video" autoplay playsinline muted></video>
```
lalli-flores: Shouldn't `muted` be on the audio track? Or does the video have its own audio track?

PR #51, opened 2020-01-28
fix(MeetingsSdkAdapter): remove finalize from getMeeting()

akoushke:


src/MeetingsSDKAdapter.js
```diff
@@ -1,6 +1,6 @@
 import {MeetingsAdapter, MeetingControlState} from '@webex/component-adapter-interfaces';
-import {concat, from, fromEvent, Observable} from 'rxjs';
-import {filter, finalize, map, merge, publishReplay, refCount} from 'rxjs/operators';
+import {concat, from, fromEvent, merge, Observable} from 'rxjs';
```
lalli-flores: Can't believe we were importing this wrong!

PR #54, opened 2020-01-28
Complete meeting

lalli-flores:
Plus control names match component default


src/MeetingsSDKAdapter.js
```diff
@@ -131,11 +131,13 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
    *
-   * @param {string} ID  ID of the meeting to fetch
+   * @param {string} ID     ID of the meeting to fetch
+   * @param {Object} event  Event type that was triggered
+   * @param {Subject} end$  Subject used to complete @see {@link getMeeting}
    * @memberof MeetingsSDKAdapter
    * @private
    */
-  removeMedia(ID, {type}) {
+  removeMedia(ID, {type}, end$) {
```
akoushke: no switch case needed here anymore. for now, `MEDIA:STOPPED` gets only triggered if we are leaving the meeting. Therefore, we should remove all attributes from the existing meeting object except the ID. to just keep record of it. NOTE: the ID is SDK MEETING ID which will be removed after leaving the meeting from the meetings collections

scripts/puppeteer.server.js
```diff
@@ -19,13 +19,18 @@ function handleVideo() {
...
 function getMeeting() {
-  webexSDKAdapter.meetingsAdapter.getMeeting(MEETING_ID).subscribe((meeting) => {
-    document.getElementById('remote-audio').srcObject = meeting.remoteAudio;
-    document.getElementById('remote-video').srcObject = meeting.remoteVideo;
-    document.getElementById('local-audio').srcObject = meeting.localAudio;
-    document.getElementById('local-video').srcObject = meeting.localVideo;
-    document.getElementById('meeting-title').innerHTML = meeting.title;
-  });
+  webexSDKAdapter.meetingsAdapter.getMeeting(MEETING_ID).subscribe(
+    (meeting) => {
```
akoushke: can you log out the meeting object here as well?

src/MeetingsSDKAdapter.js
```diff
@@ -7,8 +7,8 @@ const EVENT_MEDIA_STOPPED = 'media:stopped';
 const EVENT_MEDIA_LOCAL_UPDATE = 'adapter:media:local:update';
```
akoushke: the test cases are not updated.

scripts/puppeteer.server.js
```diff
@@ -83,7 +83,7 @@ document.getElementById('actions').addEventListener('click', async (event) => {
   try {
     switch (event.target.id) {
       case 'mute-audio':
-        await webexSDKAdapter.meetingsAdapter.meetingControls.audio.action(MEETING_ID);
+        await webexSDKAdapter.meetingsAdapter.meetingControls['mute-audio'].action(MEETING_ID);
```
akoushke: not part of your work but can you make the code look consistent? for both `audio` and `video` cases?

scripts/puppeteer.server.js
```diff
@@ -83,7 +83,7 @@ document.getElementById('actions').addEventListener('click', async (event) => {
...
         break;
       case 'mute-video':
         await webexSDKAdapter.meetingsAdapter.handleLocalVideo(MEETING_ID);
```
akoushke: talking about this line and line 86

src/MeetingsSDKAdapter.js
lalli-flores: Removed switch case

PR #56, opened 2020-01-29
Fetch Title correctly

akoushke:


src/MeetingsSDKAdapter.js
```diff
@@ -161,6 +161,7 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
    */
   async fetchMeetingTitle(destination) {
     const {id, type} = deconstructHydraId(destination);
+
```
lalli-flores: Can you remove this new line?

src/MeetingsSDKAdapter.js
```diff
@@ -196,7 +197,7 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
    */
   createMeeting(target) {
     return from(this.datasource.meetings.create(target)).pipe(
-      flatMap(({id, sipUri}) => from(this.fetchMeetingTitle(sipUri)).pipe(map((title) => ({ID: id, title})))),
+      flatMap(({id}) => from(this.fetchMeetingTitle(target)).pipe(map((title) => ({ID: id, title})))),
```
lalli-flores: Can you rename the argument from `target` to `destination`?
Previously we couldn't do that because we were destructing the attribute from SDK meeting object with the name, but that's no longer the case. I think it makes it more readable as it is consistent with the wording in other places.

rollup.config.js
```diff
@@ -36,11 +37,7 @@ export default [
       babel({
         runtimeHelpers: true,
       }),
-      commonJS({
-        // TODO: Remove workaround once fixed in SDK
-        // explicitly specify unresolvable named exports
-        namedExports: {'@webex/common': ['deconstructHydraId']},
-      }),
+      commonJS(),
```
lalli-flores: Glad we can remove this workaround 💯

lalli-flores:
Don't forget to also change the error message :)

PR #57, opened 2020-02-04
getMe() and getPerson() handle users with no presence

lalli-flores:
Whenever a user did not have presence enabled, the observables returned from `getMe()` and `getPerson()` would error out.
This PR handles users with a disabled presence (status in Teams).
In practice, this makes the observable error out in the components and also complete the observables when they were expected to be kept alive.

### DIFF
--- a/src/PeopleSDKAdapter.js
+++ b/src/PeopleSDKAdapter.js
@@ -65,10 +65,14 @@ export default class PeopleSDKAdapter extends PeopleAdapter {
    * @memberof PeopleSDKAdapter
    */
   getMe() {
+    // Get person data of the current access token bearer
     return defer(() => this.fetchPerson('me')).pipe(
       flatMap((person) =>
+        // Get person status information from presence plug-in
         defer(() => this.datasource.internal.presence.get([person.id])).pipe(
-          catchError(() => of({status: null})), // When SDK throws error, don't set a status
+          // When SDK throws error, don't set a status
+          catchError(() => of({status: null})),
+          // Combine person data and presence data to send back
           map(({status}) => ({...person, status: this.getStatus(status)}))
         )
       )
--- a/src/PeopleSDKAdapter.js
+++ b/src/PeopleSDKAdapter.js
@@ -1,5 +1,5 @@
-import {concat, from, fromEvent} from 'rxjs';
-import {flatMap, filter, finalize, map, publishReplay, refCount, take} from 'rxjs/operators';
+import {concat, defer, from, fromEvent, of} from 'rxjs';
+import {catchError, flatMap, filter, finalize, map, publishReplay, refCount} from 'rxjs/operators';
 import {deconstructHydraId} from '@webex/common';
 import {PeopleAdapter, PersonStatus} from '@webex/component-adapter-interfaces';
 
@@ -65,13 +65,13 @@ export default class PeopleSDKAdapter extends PeopleAdapter {
    * @memberof PeopleSDKAdapter
    */
   getMe() {
-    return from(this.fetchPerson('me')).pipe(
+    return defer(() => this.fetchPerson('me')).pipe(
       flatMap((person) =>
-        from(this.datasource.internal.presence.get([person.id])).pipe(
+        defer(() => this.datasource.internal.presence.get([person.id])).pipe(
+          catchError(() => of({status: null})), // When SDK throws error, don't set a status
           map(({status}) => ({...person, status: this.getStatus(status)}))
         )
-      ),
-      take(1)
+      )
     );
   }
 
--- a/src/PeopleSDKAdapter.js
+++ b/src/PeopleSDKAdapter.js
@@ -1,4 +1,4 @@
-import {concat, defer, from, fromEvent, of} from 'rxjs';
+import {concat, defer, fromEvent, of} from 'rxjs';
 import {catchError, flatMap, filter, finalize, map, publishReplay, refCount} from 'rxjs/operators';
 import {deconstructHydraId} from '@webex/common';
 import {PeopleAdapter, PersonStatus} from '@webex/component-adapter-interfaces';
@@ -90,12 +90,13 @@ export default class PeopleSDKAdapter extends PeopleAdapter {
   getPerson(ID) {
     if (!(ID in this.getPersonObservables)) {
       const personUUID = deconstructHydraId(ID).id;
-      const person$ = from(this.fetchPerson(ID));
+      const person$ = defer(() => this.fetchPerson(ID));
 
-      // Subscribe to `Apheleia` internal service to listen for status changes
-      // And update the person object with status response from the subscription
-      const personWithStatus$ = from(this.datasource.internal.presence.subscribe(personUUID)).pipe(
-        map((data) => data.responses[0].status.status),
+      // Subscribe to 'Apheleia' internal service to listen for status changes
+      // Update the Person object with status response from the subscription
+      const personWithStatus$ = defer(() => this.datasource.internal.presence.subscribe(personUUID)).pipe(
+        map((data) => data.responses[0].status.status), // This returns only the status data from subscription
+        catchError(() => of(null)), // If subscription fails, don't set a status
         flatMap((status) => person$.pipe(map((person) => ({...person, status: this.getStatus(status)}))))
       );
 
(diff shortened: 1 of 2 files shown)


akoushke:
it would have been better and easier to follow the changes, if you had a different commit for utilizing `defer`

lalli-flores:
> looks like the subscription is not happening anymore. Not sure if you fully had tested this change, but i'm not receiving any data after I change the status of my account from `ACTIVE` to `DND` or vise versa.

Yes, I tested my changes. I had no issues receiving data, _even when_ I started with presence turned off.

src/PeopleSDKAdapter.js
```diff
@@ -62,13 +65,13 @@ export default class PeopleSDKAdapter extends PeopleAdapter {
    * @memberof PeopleSDKAdapter
    */
   getMe() {
-    return from(this.fetchPerson('me')).pipe(
+    return defer(() => this.fetchPerson('me')).pipe(
```
adamweeks: What benefit does using `defer` here give us? Could you add some comments around why things are done this way? I'm sure other contributors will look to these functions as reference for new ones being created and it will help to understand the reasoning since the rxjs stuff is still brand new to most of us.

src/PeopleSDKAdapter.test.js
```diff
@@ -28,14 +32,39 @@ describe('People SDK Adapter', () => {
...
+
+      peopleSDKAdapter.getMe().subscribe((person) => {
+        expect(person).toMatchObject({
+          status: null,
+        });
+        done();
+      });
+    });
+
+    test('completes after one emission', (done) => {
```
adamweeks: Why does `getMe` complete? Wouldn't we want to get presence status updates for "me" as well?

akoushke:
> > looks like the subscription is not happening anymore. Not sure if you fully had tested this change, but i'm not receiving any data after I change the status of my account from `ACTIVE` to `DND` or vise versa.
> 
> Yes, I tested my changes. I had no issues receiving data, _even when_ I started with presence turned off.

this is the code that I had in my `start.js` file checked out to your brach. I'm not getting any data after a status change from the native client.

```js
import * as dotenv from 'dotenv';
import Webex from 'webex';

import WebexSDKAdapter from '../src/WebexSDKAdapter';

// Allows the scripts to read the corresponding configuration
dotenv.config({path: `${__dirname}/../.env.development`});

if (!process.env.WEBEX_ACCESS_TOKEN) {
  // eslint-disable-next-line no-console
  console.error('Please set a valid WEBEX_ACCESS_TOKEN environment in `env.development` file at the root level.');

  process.exit(1);
}

const webex = new Webex({
  credentials: process.env.WEBEX_ACCESS_TOKEN,
});

// eslint-disable-next-line no-unused-vars
const webexSDKAdapter = new WebexSDKAdapter(webex);

setTimeout(async () => {
  await webexSDKAdapter.connect();

  // webexSDKAdapter.peopleAdapter.getMe().subscribe((person) => console.log(person));
  webexSDKAdapter.peopleAdapter
    .getPerson('Y2lzY29zcGFyazovL3VzL1BFT1BMRS80N2MzMmQwYi0wNDQ0LTQ2MGQtOGJjZS0yMjY1YjUwMWFhYzU')
    .subscribe((person) => console.log(person));
}, 0);


export default webexSDKAdapter;
// --------------------------------------------------------------------------------
// Add your test/development code here

```

src/PeopleSDKAdapter.test.js
akoushke: this was the design that we agreed on while ago, the goal was to fetch the data one time. if the use needs to listen to the status change, they can invoke `getPerson(myID)` now.

src/PeopleSDKAdapter.js
akoushke: form what I learned, `defer` triggers the action only if there is a subscription. this makes the code a bit more performant. In our case, `defer` resolves the promise only if there is a subscription.  **However**, I don't know this would be much beneficial in our case. where we always subscribe to the function that returns an observable. Thats means `defer` does what `from` does essentially. resolving the promise right after invoking the parent function.

lalli-flores: There is a good explanation on this [StackOverflow answer on `defer` vs `from`](https://stackoverflow.com/questions/39319279/convert-promise-to-observable/54111045#54111045).

In summary, the main difference between `defer` and `from` is **when** the promise gets executed. With `from`, the promise gets executed right away, even when the observable is not subscribed to (this is one of the reasons that we would get "weird" executions flows when debugging), and because of that, the promise only gets resolved once. This means that even when we subscribe later in the code to the same `from` observable, we would still get old values because it will returned whatever the promised had resolved to previously.
With `defer`, the promise gets executed on subscription, so every new subscription to `defer` will execute the promise again, returning a fresh value. So using `defer` will allow us to always get fresh data in-between subscriptions, and it also works more inline with the mental flow of the code

src/PeopleSDKAdapter.test.js
lalli-flores: What @akoushke said 😛

src/PeopleSDKAdapter.js
akoushke: shouldn't this change be applied to all of our adapters then? perhaps a different PR/JIRA? I don't personally think we should have it here, as this is more `perf` work than a `fix`

lalli-flores:
> > > looks like the subscription is not happening anymore. Not sure if you fully had tested this change, but i'm not receiving any data after I change the status of my account from `ACTIVE` to `DND` or vise versa.
> > 
> > 
> > Yes, I tested my changes. I had no issues receiving data, _even when_ I started with presence turned off.
> 
> this is the code that I had in my `start.js` file checked out to your brach. I'm not getting any data after a status change from the native client.
> 
> ```js
> import * as dotenv from 'dotenv';
> import Webex from 'webex';
> 
> import WebexSDKAdapter from '../src/WebexSDKAdapter';
> 
> // Allows the scripts to read the corresponding configuration
> dotenv.config({path: `${__dirname}/../.env.development`});
> 
> if (!process.env.WEBEX_ACCESS_TOKEN) {
>   // eslint-disable-next-line no-console
>   console.error('Please set a valid WEBEX_ACCESS_TOKEN environment in `env.development` file at the root level.');
> 
>   process.exit(1);
> }
> 
> const webex = new Webex({
>   credentials: process.env.WEBEX_ACCESS_TOKEN,
> });
> 
> // eslint-disable-next-line no-unused-vars
> const webexSDKAdapter = new WebexSDKAdapter(webex);
> 
> setTimeout(async () => {
>   await webexSDKAdapter.connect();
> 
>   // webexSDKAdapter.peopleAdapter.getMe().subscribe((person) => console.log(person));
>   webexSDKAdapter.peopleAdapter
>     .getPerson('Y2lzY29zcGFyazovL3VzL1BFT1BMRS80N2MzMmQwYi0wNDQ0LTQ2MGQtOGJjZS0yMjY1YjUwMWFhYzU')
>     .subscribe((person) => console.log(person));
> }, 0);
> 
> 
> export default webexSDKAdapter;
> // --------------------------------------------------------------------------------
> // Add your test/development code here
> ```

Ran your code, but with my user ID so I can manipulate my status, and it works fine.
Screen Shot 2020-02-05 at 12 03 51 PM

src/PeopleSDKAdapter.js
lalli-flores: Sure, it can but I don't think the work needs to be removed just because it's not everywhere else, if we know this is the right thing to do

akoushke: to me this is still a `perf` not a `fix`

lalli-flores: It is a `fix` because it is fixing the error handling when a user has presence turned off. When changing `defer` in other PRs it can be a `perf`

akoushke: `defer` is not a `fix`.

akoushke:
> 
> Ran your code, but with my user ID so I can manipulate my status, and it works fine.
> Screen Shot 2020-02-05 at 12 03 51 PM

this is super weird. I'm still not getting any data after a changing my status. 
Screen Shot 2020-02-05 at 12 31 44 PM

src/PeopleSDKAdapter.js
adamweeks: Sounds good, thanks for the explanation! We've got lots of comments in the `getPerson` function, could you add the same here?

lalli-flores: Add why using `defer`? I'm not sure that's necessary. It's a pattern for converting promises => Observables, and we didn't comment on the usage of `from`, because of the same

akoushke:
it looks like the subscription will not happen if the users are not in a same organizations. therefore, I was not getting anymore data.

lalli-flores:
> it looks like the subscription will not happen if the users are not in a same organizations. therefore, I was not getting anymore data.

So with this fix there are two scenarios covered:
* User has no presence enabled
* User is trying to access presence of another user in a different org

PR #61, opened 2020-02-27
Exclude webex from the bundle

akoushke:


rollup.config.js
```diff
@@ -11,54 +9,23 @@ const output = (name, format) => ({
...
-    'utf-8-validate': 'utf8Validate',
+    webex: 'webex',
   },
 });
 
 export default [
   {
     input: 'src/index.js',
-    output: [
-      output('webexSDKComponentAdapter', 'cjs'),
```
adamweeks: Removing this will break the package.json's `"main": "dist/webexSDKComponentAdapter.cjs.js",`

akoushke:
> Can you update your commit types so that they don't trigger a release?
> The Rollup changes could use `refactor` and `package.json` could use `build`.
> 
> I would also suggest a bit more specific comment on rollup commits, like `removing webex-related configurations`

I need to release this to use the bundle for https://github.com/webex/widgets/pull/10

PR #64, opened 2020-05-21
Add Screen Sharing Meeting Control

ZixiXu:
To implement the sharing feature, I added ```handleLocalShare``` and ```shareControl``` to connect to webex-js-sdk to start/stop sharing, and emit sharing status to upper UI level.

### DIFF
--- a/scripts/puppeteer.server.js
+++ b/scripts/puppeteer.server.js
@@ -18,6 +18,16 @@ function handleVideo() {
   });
 }
 
+function handleShare() {
+  webexSDKAdapter.meetingsAdapter.meetingControls['share-screen']
+    .display(MEETING_ID)
+    .subscribe((data) => {
+    const startShare = document.getElementById('share-screen');
+
+    startShare.innerHTML = data.tooltip;
+  });
+}
+
 function getMeeting() {
   webexSDKAdapter.meetingsAdapter.getMeeting(MEETING_ID).subscribe(
     (meeting) => {
@@ -28,6 +38,8 @@ function getMeeting() {
       document.getElementById('remote-video').srcObject = meeting.remoteVideo;
       document.getElementById('local-audio').srcObject = meeting.localAudio;
       document.getElementById('local-video').srcObject = meeting.localVideo;
+      document.getElementById('local-share').srcObject = meeting.localShare;
+      document.getElementById('remote-share').srcObject = meeting.remoteShare;
       document.getElementById('meeting-title').innerHTML = meeting.title;
     },
     () => {},
@@ -67,6 +79,7 @@ document.getElementById('dialer').addEventListener('click', async (event) => {
           getMeeting();
           handleAudio();
           handleVideo();
+          handleShare();
         });
         break;
       case 'join-meeting':
@@ -92,6 +105,9 @@ document.getElementById('actions').addEventListener('click', async (event) => {
         break;
       case 'mute-video':
         await webexSDKAdapter.meetingsAdapter.meetingControls['mute-video'].action(MEETING_ID);
+        break;
+      case 'share-screen':
+        await webexSDKAdapter.meetingsAdapter.meetingControls['share-screen'].action(MEETING_ID);
         break;
     }
   } catch (error) {
--- a/src/MeetingsSDKAdapter.js
+++ b/src/MeetingsSDKAdapter.js
@@ -80,26 +80,48 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
   }
 
   /**
-   * Returns the local device media
+   * Returns a promise to the local device media streams.
    *
    * @param {string} ID  ID to retrieve the SDK meeting object to add the local media to
    * @returns {Object}
    */
   async getLocalMedia(ID) {
-    const sdkMeeting = this.fetchMeeting(ID);
     const localMedia = {localAudio: null, localVideo: null};
 
-    try {
-      const [localStream] = await sdkMeeting.getMediaStreams(DEFAULT_MEDIA_SETTINGS);
+    if (DEFAULT_MEDIA_SETTINGS.sendAudio) {
+      localMedia.localAudio = await this.getStream(ID, {sendAudio: true});
+    }
 
-      localMedia.localAudio = new MediaStream(localStream.getAudioTracks());
-      localMedia.localVideo = new MediaStream(localStream.getVideoTracks());
-    } catch (error) {
-      // eslint-disable-next-line no-console
-      console.error(`Unable to retrieve local stream media "${ID}"`, error);
+    if (DEFAULT_MEDIA_SETTINGS.sendVideo) {
+      localMedia.localVideo = await this.getStream(ID, {sendVideo: true});
     }
 
     return localMedia;
+  }
+
+  /**
+   * Returns a promise to a local media stream based on the given constraints.
+   * @see {@link MediaStream|https://developer.mozilla.org/en-US/docs/Web/API/MediaStream}.
+   *
+   * @param {string} ID  ID of the meeting for which to fetch streams
+   * @param {Object} constraint
+   * @returns {Promise.<MediaStream>}  Requested media stream
+   * @memberof MeetingsSDKAdapter
+   * @private
+   */
+  async getStream(ID, constraint) {
+    let localStream = null;
+
+    try {
+      const sdkMeeting = this.fetchMeeting(ID);
+
+      [localStream] = await sdkMeeting.getMediaStreams(constraint);
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error(`Unable to retrieve local media stream for meeting "${ID}" with constraint "${constraint}"`, error);
+    }
+
+    return localStream;
   }
 
   /**
@@ -156,7 +178,7 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
    * Supported destinations are person ID, room ID and SIP URI.
    *
    * @param {string} destination  Virtual meeting destination
-   * @returns {Promise<string>}
+   * @returns {Promise.<string>}
    * @memberof MeetingsSDKAdapter
    */
   async fetchMeetingTitle(destination) {
@@ -246,7 +268,19 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
     try {
       const sdkMeeting = this.fetchMeeting(ID);
       const {localAudio, localVideo} = this.meetings[ID];
-      const [localStream] = await sdkMeeting.getMediaStreams(DEFAULT_MEDIA_SETTINGS);
+      const localStream = new MediaStream();
+
+      if (localAudio) {
+        const tracks = localAudio.getTracks();
+
+        tracks.forEach((track) => localStream.addTrack(track));
+      }
+
+      if (localVideo) {
+        const tracks = localVideo.getTracks();
+
+        tracks.forEach((track) => localStream.addTrack(track));
+      }
 
       await sdkMeeting.join();
 
--- a/src/MeetingsSDKAdapter.js
+++ b/src/MeetingsSDKAdapter.js
@@ -157,6 +157,22 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
   }
 
   /**
+   * Stops the tracks of the given media stream.
+   * @see {@link MediaStream|https://developer.mozilla.org/en-US/docs/Web/API/MediaStream}.
+   *
+   * @param {MediaStream} stream  media stream for which to stop tracks
+   * @memberof MeetingsSDKAdapter
+   * @private
+   */
+  stopStream(stream) {
+    if (stream) {
+      const tracks = stream.getTracks();
+
+      tracks.forEach((track) => track.stop());
+    }
+  }
+
+  /**
    * Update the meeting object by removing all media.
    *
    * @param {string} ID  ID of the meeting to fetch
@@ -164,6 +180,12 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
    * @private
    */
   removeMedia(ID) {
+    if (this.meetings && this.meetings[ID]) {
+      this.stopStream(this.meetings[ID].localAudio);
+      this.stopStream(this.meetings[ID].localVideo);
+      this.stopStream(this.meetings[ID].localShare);
+    }
+
     this.meetings[ID] = {
       ...this.meetings[ID],
       localAudio: null,
--- a/src/MeetingsSDKAdapter.js
+++ b/src/MeetingsSDKAdapter.js
@@ -1,29 +1,49 @@
 import {MeetingsAdapter, MeetingControlState} from '@webex/component-adapter-interfaces';
 import {deconstructHydraId} from '@webex/common';
-import {concat, from, fromEvent, merge, Observable, Subject} from 'rxjs';
+import {concat, defer, from, fromEvent, merge, Observable, Subject} from 'rxjs';
 import {flatMap, filter, map, publishReplay, refCount, takeUntil, tap} from 'rxjs/operators';
 
+// JS SDK Events
 const EVENT_MEDIA_READY = 'media:ready';
 const EVENT_MEDIA_STOPPED = 'media:stopped';
+const EVENT_LOCAL_SHARE_STOP = 'meeting:stoppedSharingLocal';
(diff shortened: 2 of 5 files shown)


scripts/puppeteer.server.js
```diff
@@ -93,6 +110,9 @@ document.getElementById('actions').addEventListener('click', async (event) => {
       case 'mute-video':
         await webexSDKAdapter.meetingsAdapter.meetingControls['mute-video'].action(MEETING_ID);
         break;
+      case 'share-control':
```
lalli-flores, line 113:
```suggestion
      case 'share-screen':
```

src/MeetingsSDKAdapter.js
```diff
@@ -86,20 +103,40 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
    */
   async getLocalMedia(ID) {
-    const sdkMeeting = this.fetchMeeting(ID);
     const localMedia = {localAudio: null, localVideo: null};
 
+    if (DEFAULT_MEDIA_SETTINGS.sendAudio) {
+      localMedia.localAudio = await this.getStream(ID, {sendAudio: true});
+    }
+    if (DEFAULT_MEDIA_SETTINGS.sendVideo) {
+      localMedia.localVideo = await this.getStream(ID, {sendVideo: true});
```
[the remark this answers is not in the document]
ZixiXu: This is to handle the case where user blocks his/her camera but allow mic, or vise versa. Browser can still get the mic or cam separately after doing multiple calls. We can discuss this in a meeting :)

src/MeetingsSDKAdapter.js
```diff
@@ -86,20 +103,40 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
     try {
-      const [localStream] = await sdkMeeting.getMediaStreams(DEFAULT_MEDIA_SETTINGS);
+      const sdkMeeting = this.fetchMeeting(ID);
+
+      if (sdkMeeting) {
+        const [localStream] = await sdkMeeting.getMediaStreams(constraint);
 
-      localMedia.localAudio = new MediaStream(localStream.getAudioTracks());
-      localMedia.localVideo = new MediaStream(localStream.getVideoTracks());
+        return localStream;
```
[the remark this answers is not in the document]
ZixiXu: I see. I can keep the return at the end. Thanks!

src/MeetingsSDKAdapter.js
```diff
@@ -129,11 +166,34 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
+  /**
+   * Stop the given media stream.
+   *
+   * @param {stream} stream  stream of the meeting to stop
+   * @memberof MeetingsSDKAdapter
+   * @private
+   */
+  stopStream(stream) {
+    if (stream) {
+      const tracks = stream.getTracks && stream.getTracks();
```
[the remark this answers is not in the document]
ZixiXu: Yes it is a member of API. It is just a sanity check for the stream, I can remove it.

src/MeetingsSDKAdapter.js
```diff
@@ -142,12 +202,19 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
    * @private
    */
   removeMedia(ID) {
+    if (this.meetings && this.meetings[ID]) {
+      this.stopStream(this.meetings[ID].localAudio);
+      this.stopStream(this.meetings[ID].localVideo);
+      this.stopStream(this.meetings[ID].localShare);
+    }
```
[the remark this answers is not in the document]
ZixiXu: No problem! :)

src/MeetingsSDKAdapter.js
```diff
@@ -129,11 +166,34 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
+        break;
       default:
         break;
     }
   }
 
+  /**
+   * Stop the given media stream.
+   *
+   * @param {stream} stream  stream of the meeting to stop
```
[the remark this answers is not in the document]
ZixiXu: Yes, should be a MediaStream

ZixiXu:
Thanks @lalli-flores ! I have updated the code according to your review.

src/MeetingsSDKAdapter.js
```diff
@@ -509,6 +590,157 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
+      sdkMeeting.emit(EVENT_MEDIA_LOCAL_UPDATE, {
+        control: SHARE_CONTROL,
+        state: MeetingControlState.DISABLED,
+      });
+
+      await sdkMeeting.updateShare({stream: response[1], sendShare: true, receiveShare: true});
+    };
+
+    const disableSharingStream = async () => {
+      sdkMeeting.emit(EVENT_MEDIA_LOCAL_UPDATE, {
```
lalli-flores, line 638: I think you should emit only after the share stream was successfully updated and stopped, instead of the beginning

src/MeetingsSDKAdapter.js
```diff
@@ -509,6 +590,157 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
+      await sdkMeeting.updateShare({
+        sendShare: false,
+        receiveShare: true,
+      });
+
+      this.stopStream(this.meetings[ID].localShare);
+      this.meetings[ID].localShare = null;
+    };
+
+    const resetSharingStream = () => {
```
lalli-flores, line 652: We should probably propagate the original error
```suggestion
    const resetSharingStream = (error) => {
```

src/MeetingsSDKAdapter.js
```diff
@@ -509,6 +590,157 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
+      // eslint-disable-next-line no-console
+      console.error(`Unable to update local share for meeting "${ID}"`);
```
lalli-flores, line 654: We should probably propagate the original error
```suggestion
      console.error(`Unable to update local share for meeting "${ID}"`, error);
```

src/MeetingsSDKAdapter.js
```diff
@@ -509,6 +590,157 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
+      sdkMeeting.emit(EVENT_MEDIA_LOCAL_UPDATE, {
```
lalli-flores, line 655: Emitting the event should be the last thing in the flow

src/MeetingsSDKAdapter.js
```diff
@@ -509,6 +590,157 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
+        control: SHARE_CONTROL,
+        state: MeetingControlState.INACTIVE,
+      });
+
+      if (this.meetings[ID].localShare) {
+        // eslint-disable-next-line no-console
+        console.warn(`Stop local share stream for meeting "${ID}"`);
```
lalli-flores, line 662:
```suggestion
        console.warn(`Stopping local share stream for meeting "${ID}"`);
```

src/MeetingsSDKAdapter.js
```diff
@@ -509,6 +590,157 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
+      map(({state}) => {
+        switch (state) {
+          case MeetingControlState.DISABLED:
+            return shareBusyMsg;
+          case MeetingControlState.INACTIVE:
+            return shareInactiveMsg;
+          case MeetingControlState.ACTIVE:
+            return shareActiveMsg;
+          default:
+            return null;
```
lalli-flores, line 736: I think the control should be disabled by default (e.g. going into some unknown state)

src/MeetingsSDKAdapter.js
```diff
@@ -509,6 +590,157 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
+        }
+      })
```
lalli-flores, line 738: Can we follow the one-exit control pattern?
```suggestion
        return eventData;
      })
```

where the switch case would set `eventData`

src/MeetingsSDKAdapter.js
```diff
@@ -535,6 +767,30 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
         map((event) => this.attachMedia(ID, event))
       );
 
+      const meetingWithMediaStoppedSharingLocalEvent$ = fromEvent(sdkMeeting, EVENT_MEETING_STOPPED_SHARING_LOCAL).pipe(
+        // eslint-disable-next-line no-console
+        map(() => console.log('EVENT_MEETING_STOPPED_SHARING_LOCAL is triggered', this)),
```
lalli-flores, line 772: Can you elaborate on what `meetingWithMediaStoppedSharingLocalEvent$` and `meetingWithMediaStartedSharingLocalEvent$` are trying to accomplish?
I feel that this is misplaced.

src/MeetingsSDKAdapter.js
```diff
@@ -5,21 +5,32 @@ import {flatMap, filter, map, publishReplay, refCount, takeUntil, tap} from 'rxj
...
+  MEDIA_TYPE_REMOTE_AUDIO,
+  MEDIA_TYPE_REMOTE_VIDEO,
+  MEDIA_TYPE_LOCAL_SHARE,
+  MEDIA_TYPE_REMOTE_SHARE,
+];
 const DEFAULT_MEDIA_SETTINGS = {
   receiveVideo: true,
-  receiveAudio: true,
-  receiveShare: false,
+  receiveAudio: false,
```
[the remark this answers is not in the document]
ZixiXu: Please refer to getLocalMedia(), I still use the constraints (sendAudio, not receiveAudio) from DEFAULT_MEDIA_SETTINGS. And DEFAULT_MEDIA_SETTINGS is also used in joinMeeting().

src/MeetingsSDKAdapter.js
```diff
@@ -509,6 +590,157 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
+   * @param {string} ID ID of the meeting to start/stop sharing
+   * @memberof MeetingJSONAdapter
+   */
+  async handleLocalShare(ID) {
+    const sdkMeeting = this.fetchMeeting(ID);
+
+    if (!sdkMeeting) {
+      // eslint-disable-next-line no-console
+      console.error(`
+        Unable to update share for meeting "${ID}": sdk meeting does not exist
```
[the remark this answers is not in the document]
ZixiXu: ok, I will remove the check

src/MeetingsSDKAdapter.js
ZixiXu: This is a workaround for a bug of async updateShare() call. The async call may need ~30s to finish execution, or even cannot finish execution. If the emit is called after the async call, the user will never see the state change of UI and get confused. So we moved this emit call to before the async call.

src/MeetingsSDKAdapter.js
ZixiXu: ok, since it is not async call, I can move it

src/MeetingsSDKAdapter.js
ZixiXu: So you think it is better to return the disabledShare object for default?

src/MeetingsSDKAdapter.js
ZixiXu: These are similar to meetingWithMediaStoppedEvent$ in getMeeting().  

meetingWithMediaStartedSharingLocalEvent$ is the event when local sharing stream is started and added to SDP offer.
meetingWithMediaStoppedSharingLocalEvent$ is the event monitoring the ```onended``` event of the local sharing stream. If the user clicks the "stop share" button (which is native button of browser), this event will help to detect that the user wants to stop share.
Both events take effects during the whole meeting period, just like meetingWithMediaStoppedEvent$.

Screen Shot 2020-06-03 at 11 57 33 AM


Maybe a better name for this event to increase the readability?

ZixiXu:
Thanks @lalli-flores! I have updated the code for the second round :)

src/MeetingsSDKAdapter.js
chrisadubois: We should probably mark it with a TODO and a jira

src/MeetingsSDKAdapter.js
```diff
@@ -509,6 +588,154 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
+        receiveShare: true,
+      });
+
+      this.stopStream(this.meetings[ID].localShare);
+      this.meetings[ID].localShare = null;
+    };
+
+    const resetSharingStream = (error) => {
+      // eslint-disable-next-line no-console
+      console.error(`Unable to update local share for meeting "${ID}"`, error);
```
chrisadubois, line 640: We should change this to the SDK logger. the component should be logging it's own debuggability information

src/MeetingsSDKAdapter.js
```diff
@@ -535,6 +762,30 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
 
+      const meetingWithMediaStoppedSharingLocalEvent$ = fromEvent(sdkMeeting, EVENT_MEETING_STOPPED_SHARING_LOCAL).pipe(
+        // eslint-disable-next-line no-console
+        map(() => console.log('EVENT_MEETING_STOPPED_SHARING_LOCAL is triggered', this)),
+        tap(async () => {
+          try {
+            await this.meetingControls[SHARE_CONTROL].action(ID);
+          } catch (error) {
+            // eslint-disable-next-line no-console
+            console.error('Unable to perform an action for EVENT_MEETING_STOPPED_SHARING_LOCAL:', error);
```
chrisadubois, line 773: Same here

src/MeetingsSDKAdapter.js
ZixiXu: Reported as this jira: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-148910

src/MeetingsSDKAdapter.js
ZixiXu: ok, just updated. Thanks!

src/MeetingsSDKAdapter.js
lalli-flores: Thanks for reporting it 👍🏻

src/MeetingsSDKAdapter.js
lalli-flores: Yes because if for whatever reason the code ends up in the `default` case, that means something we didn't expect emitted the event and we potentially won't know how to handle the flow. Therefore we should allow the user to use the share control since we won't know what could happen (even potentially crash).

src/MeetingsSDKAdapter.js
lalli-flores: I think the logic here is not right. Whatever happens in `getMeeting` should result in the final observable emitting a `Meeting` object as defined by the component system data model. The resulting emission should have the desired state.
For instance, if a new local share stream is added, then the `Meeting` object should have that stream. If the streamed is stopped, then it should be removed, etc.
The actions are invoked by the component when the user clicks the control and should not be invoked directly in the adapter.

src/MeetingsSDKAdapter.js
```diff
@@ -509,6 +588,181 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
+        return eventData;
+      })
+    );
+
+    const meetingWithMediaStoppedSharingLocalEvent$ = fromEvent(sdkMeeting, EVENT_LOCAL_SHARE_STOP).pipe(
+      // eslint-disable-next-line no-console
+      tap(() => console.log('EVENT_LOCAL_SHARE_STOP is triggered', this)),
+      tap(async () => {
+        try {
+          await this.meetingControls[SHARE_CONTROL].action(ID);
```
lalli-flores, line 743: Why are you invoking the control's action here?

ZixiXu: sdk will handle the stop event, but it only stops the local share stream, we need to send message to let locus know that the sender now stop the sharing, so we called the action function
I have created the jira here: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-152004

lalli-flores: Ok, then `tap` is the wrong operator to use here. We need to use a combination of `flatMap` and `defer` to be able to add the promise as part of the observable stream flow.
```suggestion
          flatMap(defer(() => this.meetingControls[SHARE_CONTROL].action(ID)));
```

PR #65, opened 2020-07-08
fix(share): use correct key value for media settings

stevem63:
The JS SDK's `addMedia` function requires the media settings to have the key `mediaSettings`.


lalli-flores:
@stevem63, can you update the scope of the commit message to use the filename for consistency with the [changelog](https://github.com/webex/sdk-component-adapter/blob/master/CHANGELOG.md)?
```
fix(MeetingsSdkAdapter): use correct key name for media settings on addMedia

stevem63:
@lalli-flores done 👍

PR #67, opened 2020-07-09
feat(MeetingsSdkAdapter): prevent showing remote share to presenter

stevem63:


lalli-flores:
Closing as this is not the right fix for screen share streams. There is an agreed fix on the way in the SDK that will add an event for share that the adapter should listen to.

PR #69, opened 2020-07-20
fix(MeetingsSdkAdapter): mute/unmute toggle

jdaly13:
Fixes an issue where user can not mute or unmute if they join a meeting while muted
when joining a meeting localAudio can't be set to null until tracks have been added to the local stream for audio

Fixes #SPARK-[148175](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-148174)

### DIFF
--- a/src/MeetingsSDKAdapter.js
+++ b/src/MeetingsSDKAdapter.js
@@ -57,6 +57,7 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
     super(datasource);
     this.getMeetingObservables = {};
     this.meetings = {};
+    this.audioEnabled = true;
 
     this.meetingControls[JOIN_CONTROL] = {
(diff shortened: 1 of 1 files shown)


src/MeetingsSDKAdapter.js
```diff
@@ -57,6 +57,7 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
     super(datasource);
     this.getMeetingObservables = {};
     this.meetings = {};
+    this.audioEnabled = true;
```
lalli-flores, line 60: This cannot be a instance variable. The meetings adapter acts like a singleton manager for potentially multiple meetings.

jdaly13: I think I found a better solution opening up a new PR closing this out

PR #70, opened 2020-07-20
Fix Audio and Video Mute at Join

jdaly13:
Fixes an issue where user can not mute or unmute if they join a meeting when the set their audio to be muted before joining meeting
when joining a meeting localAudio can't be set to null until tracks have been added to the local stream for audio

Fixes #SPARK-[148175](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-148174)

### DIFF
--- a/src/MeetingsSDKAdapter.js
+++ b/src/MeetingsSDKAdapter.js
@@ -323,7 +323,6 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
   async joinMeeting(ID) {
     try {
       const sdkMeeting = this.fetchMeeting(ID);
-      const {localVideo} = this.meetings[ID];
       const localStream = new MediaStream();
 
       const localAudio = this.meetings[ID].localAudio || this.meetings[ID].disabledLocalAudio;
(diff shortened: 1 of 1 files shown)


jdaly13:
Essentially when you join a meeting and you are on the preview screen and click the mute button - when you join the meeting the audio track is not set so you can not unmute yourself - the button still toggles so you're muting and unmuting but there's no audio mute/unmute with

src/MeetingsSDKAdapter.js
```diff
@@ -323,14 +323,13 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
   async joinMeeting(ID) {
     try {
       const sdkMeeting = this.fetchMeeting(ID);
-      const {localAudio, localVideo} = this.meetings[ID];
+      const {localVideo} = this.meetings[ID];
```
lalli-flores, line 326: Can you do the same change for `localVideo`? Since the logic is the same, it's bound to have the same issue as audio.
This also ensure we're consistent on how we treat local streams.

jdaly13: I have updated the original commit message and updated the localVideo as well - There are some es lint formatting changes I was thinking of committing but haven't yet - let me know if I should commit those as well

lalli-flores:
@jdaly13, thanks for updating with the feedback.
What do you mean by ESLint formatting changes?

jdaly13:
@lalli-flores - when committing work I had to run 
 npm run test:eslint -- --fix 
in order to commit my changes
however I did not commit the formatting changes but it had to be run before I committed 
they are essentially just format changes but didn't want to commit them as part of bug fixes but possibly in a separate commit if desired

lalli-flores:
If they are linting changes unrelated those in this PR, no need to commit them. Seems like the [linting checks passed](https://app.circleci.com/pipelines/github/webex/sdk-component-adapter/487/workflows/8b33a85e-3b15-460a-8b3a-3dcd68b0bba0/jobs/1061/parallel-runs/0/steps/0-103) for the PR so you should be good

PR #72, opened 2020-08-14
fix(rollup): remove webex/common from external add as named export

jdaly13:
This PR is to fix [SPARK-106451](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-106451) where we do not include @webex/common as an external dependency as the current workaround.  The common JS plugin for rollup needs to have namedExports in order to find the code properly please see here
https://rollupjs.org/guide/en/#error-name-is-not-exported-by-module

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -8,7 +8,6 @@ const output = (name, format) => ({
   format,
   sourcemap: true,
   globals: {
-    '@webex/common': '@webex.common',
     rxjs: 'rxjs',
     'rxjs/operators': 'rxjs.operators',
     webex: 'webex',
@@ -20,12 +19,16 @@ export default [
     input: 'src/index.js',
     output: [output('ESMWebexSDKComponentAdapter', 'esm')],
     plugins: [
-      resolve(),
+      resolve({preferBuiltins: false}),
       babel({
         runtimeHelpers: true,
       }),
-      commonJS(),
+      commonJS({
+        namedExports: {
+          '@webex/common': ['deconstructHydraId'],
+        },
+      }),
     ],
-    external: ['@webex/common', 'rxjs', 'rxjs/operators', 'webex'],
+    external: ['rxjs', 'rxjs/operators', 'webex'],
   },
 ];


adamweeks:
Do we need to document this usages in the sdk repo?

rollup.config.js
```diff
@@ -20,12 +19,16 @@ export default [
...
       }),
-      commonJS(),
+      commonJS({
+        namedExports: {
+          '@webex/common': ['deconstructHydraId'],
+        },
+      }),
     ],
-    external: ['@webex/common', 'rxjs', 'rxjs/operators', 'webex'],
+    external: ['webex', 'rxjs', 'rxjs/operators'],
```
lalli-flores, line 32: Nit-picky but we were trying to keep the modules in alphabetical order

jdaly13:
@adamweeks I don't think so but that's not my decision.  I think this is a rollup specific issue.  Also I'm not sure if these changes are even necessary as webex/common is a dependency of webex - meaning if you define it as external in the rollup config as a workaround, you won't ever have to manually install it yourself in a 3rd party application that consumes sdk-adapter because - the sdk-adapter has webex as a peer dependency and webex will always have @webex/common.  
So in theory this "fix" is just a different workaround.

rollup.config.js
jdaly13: okay I'll update

lalli-flores:
@jdaly13 I understand what you're saying about `@webex/common` already being part of `webex` and so the function will be there in the app using the SDK adapter, however, rollup is not able to find it, and _that_ is the problem because it won't create the bundle

jdaly13:
@lalli-flores - you need to name the exports or rollup won't find them because of the way modules are exported in the sdk.  in the index file you are exporting from a particular file but not importing first so therefore the rollup commonjs helper can't find it.
https://rollupjs.org/guide/en/#error-name-is-not-exported-by-module

jdaly13:
@lalli-flores not sure what issues your having are you getting npm errors have you tried maybe doing a clean install of your packages

lalli-flores:
@jdaly13 before merging, can you squash your commits as one?
And please use the commit type `build` instead of `fix`, as we don't want tooling changes to trigger a release.

jdaly13:
commit message updated and other one squashed

PR #77, opened 2020-09-09
feat(MembershipsSdkAdapter): add meeting support

adamweeks:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-174910

### DIFF
--- a/src/MembershipsSDKAdapter.js
+++ b/src/MembershipsSDKAdapter.js
@@ -1,0 +1,109 @@
+import {BehaviorSubject, throwError} from 'rxjs';
+import {DestinationType, MembershipsAdapter} from '@webex/component-adapter-interfaces';
+
+/**
+ * Gets the active members in a meeting
+ *
+ * @private
+ * @param {object} members Members object from meeting, keyed by ID
+ * @returns {Array}
+ * @memberof MembershipsSDKAdapter
+ */
+function getActiveMembers(members) {
+  return Object.values(members)
+    .filter((member) => !member.isInMeeting)
+    .map((member) => ({
+      id: member.id,
+    }));
+}
+
+export default class MembershipsSDKAdapter extends MembershipsAdapter {
+  constructor(datasource) {
+    super(datasource);
+
+    this.membershipSubjects = {};
+  }
+
+  /**
+   * Returns an observable subject for the given membership.
+   *
+   * @private
+   * @param {string} membershipID  ID of the destination for which to get members
+   * @param {DestinationType} destinationType  type of the membership destination
+   * @returns {BehaviorSubject}
+   */
+  getOrCreateMembershipSubject(membershipID) {
+    if (!this.membershipSubjects[membershipID]) {
+      // Behavior subject will keep the last emitted object for new subscribers
+      // https://rxjs.dev/guide/subject#behaviorsubject
+      this.membershipSubjects[membershipID] = new BehaviorSubject([]);
+    }
+
+    return this.membershipSubjects[membershipID];
+  }
+
+  /**
+   * Returns an observable that emits a Membership object.
+   * Whenever there is an update to the membership, the observable
+   * will emit a new updated Membership object, if datasource permits.
+   *
+   * @param {string} destinationID  ID of the destination for which to get members
+   * @param {DestinationType} destinationType  type of the membership destination
+   * @returns {external:Observable.<Membership>} Observable stream that emits membership data
+   * @memberof MembershipAdapter
+   */
+  getMembersFromDestination(destinationID, destinationType) {
+    if (destinationType !== DestinationType.MEETING) {
+      return throwError(new Error(`getMembersFromDestination for ${destinationType} is not currently supported.`));
+    }
+
+    const meeting = this.datasource.meetings.getMeetingByType('id', destinationID);
+
+    if (!meeting) {
+      return throwError(new Error(`Meeting ${destinationID} not found.`));
+    }
+
+    const membershipID = `${destinationType}-${destinationID}`;
+
+    if (!this.membershipSubjects[membershipID]) {
+      // Behavior subject will keep the last emitted object for new subscribers
+      // https://rxjs.dev/guide/subject#behaviorsubject
+      this.membershipSubjects[membershipID] = new BehaviorSubject({
+        ID: membershipID,
+        destinationID,
+        destinationType,
+        members: [],
+      });
+    }
+
+    const membershipSubject = this.membershipSubjects[membershipID];
+
+    if (meeting.members && meeting.members.membersCollection && meeting.members.membersCollection.members) {
+      const members = getActiveMembers(meeting.members.membersCollection.members);
+
+      // First, emit the current collection
+      membershipSubject.next({
+        ID: membershipID,
+        destinationID,
+        destinationType,
+        members,
+      });
+    }
+
+    // Emit on membership updates
+    meeting.members.on('members:update', (payload) => {
+      if (payload && payload.full) {
+        const updatedMembers = getActiveMembers(payload.full);
+
+        membershipSubject.next({
+          ID: membershipID,
+          destinationID,
+          destinationType,
+          members: updatedMembers,
+        });
+      }
+    });
+
+    return membershipSubject;
+  }
+}
--- a/src/MembershipsSDKAdapter.test.js
+++ b/src/MembershipsSDKAdapter.test.js
@@ -1,0 +1,67 @@
+import {isObservable} from 'rxjs';
+import {DestinationType} from '@webex/component-adapter-interfaces';
+
+import MembershipsSDKAdapter from './MembershipsSDKAdapter';
+import createMockSDK from './__mocks__/sdk';
+
+describe('Memberships SDK Adapter', () => {
+  let membershipSDKAdapter, mockSDK, meetingID;
+
+  beforeEach(() => {
+    mockSDK = createMockSDK();
+    membershipSDKAdapter = new MembershipsSDKAdapter(mockSDK);
+    meetingID = 'meetingID';
+  });
+
+  describe('getMembersFromDestination()', () => {
+    test('returns an observable', () => {
+      expect(isObservable(membershipSDKAdapter.getMembersFromDestination())).toBeTruthy();
+    });
+
+    describe('when destination type is MEETING', () => {
+      test('emits a membership object on subscription', (done) => {
+        membershipSDKAdapter.getMembersFromDestination(meetingID, DestinationType.MEETING).subscribe((membership) => {
+          expect(membership).toMatchObject({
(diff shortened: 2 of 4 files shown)


src/MembershipsSDKAdapter.js
```diff
@@ -0,0 +1,64 @@
...
+   */
+  getMembersFromDestination(destinationID, destinationType) {
+    if (destinationType !== DestinationType.MEETING) {
+      return throwError(new Error(`getMembersFromDestination for ${destinationType} is not currently supported.`));
+    }
+
+    const meeting = this.datasource.meetings.getMeetingByType('id', destinationID);
+
+    if (!meeting) {
+      return throwError(new Error(`Meeting ${destinationID} not found in datasource.`));
```
lalli-flores, line 39:
```suggestion
      return throwError(new Error(`Meeting ${destinationID} not found.`));
```

src/MembershipsSDKAdapter.js
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    // Behavior subject will keep the last emitted object for new subscribers
+    // https://rxjs.dev/guide/subject#behaviorsubject
```
lalli-flores, line 43: Thanks for adding the comment 👍🏼

src/MembershipsSDKAdapter.test.js
```diff
@@ -0,0 +1,58 @@
...
+          expect(Array.isArray(members)).toBeTruthy();
+          done();
+        });
+      });
+
+      test('throws an error on invalid meeting ID', (done) => {
+        membershipSDKAdapter.getMembersFromDestination('badID', DestinationType.MEETING).subscribe(
+          () => {},
+          (error) => {
+            expect(error).toBeDefined();
```
lalli-flores, line 33: Can we check that is this the specific error we expect?

src/MembershipsSDKAdapter.test.js
```diff
@@ -0,0 +1,58 @@
...
+        );
+      });
+    });
+
+    describe('room destination type', () => {
+      test('throws an error on subscription', (done) => {
+        membershipSDKAdapter.getMembersFromDestination('roomID', DestinationType.ROOM).subscribe(
+          () => {},
+          (error) => {
+            expect(error).toBeDefined();
```
lalli-flores, line 45: Can we check that is this the specific error we expect?

src/MembershipsSDKAdapter.test.js
```diff
@@ -0,0 +1,58 @@
...
+    membershipSDKAdapter = new MembershipsSDKAdapter(mockSDK);
+    meetingID = 'meetingID';
+  });
+
+  describe('getMembersFromDestination()', () => {
+    test('returns an observable', () => {
+      expect(isObservable(membershipSDKAdapter.getMembersFromDestination())).toBeTruthy();
+    });
+
+    describe('meeting destination type', () => {
```
lalli-flores, line 21: Not sure the nested `describe` are necessary? it makes the test a bit confusing since it's not just about the meeting destination or the meeting type

src/MembershipsSDKAdapter.test.js
```diff
@@ -0,0 +1,58 @@
...
+          (error) => {
+            expect(error).toBeDefined();
+            done();
+          }
+        );
+      });
+    });
+
+    describe('room destination type', () => {
+      test('throws an error on subscription', (done) => {
```
lalli-flores, line 41: Removing the nested `describe`s, we would just need to update the test description to be more specific
```suggestion
      test('throws an error on invalid meeting type', (done) => {
```

src/MembershipsSDKAdapter.js
```diff
@@ -0,0 +1,64 @@
...
+    const membershipSubject = new BehaviorSubject([]);
```
lalli-flores, line 44: What is the value in making this a subject? In the current implementation whenever `getMembersFromDestination` gets called, a new subject is going to be created so there will never be multiple subscribers.
I think using the subject itself its fine, but how are you making sure that whenever there is a call for the same ID + Type combo the same subject is reused?

lalli-flores:
Maybe your commit message can be a bit more descriptive?
``
feat(MembershipsSdkAdapter): add ability to find meeting memberships

src/MembershipsSDKAdapter.js
```diff
@@ -0,0 +1,64 @@
...
+export default class MembershipsSDKAdapter extends MembershipsAdapter {
+  /**
+   * Gets the active members in a meeting
+   *
+   * @static
+   * @param {object} members Members object from meeting, keyed by ID
+   * @returns {Array}
+   * @memberof MembershipsSDKAdapter
+   */
+  static getActiveMembers(members) {
```
lalli-flores, line 13: With Babel we should be able to use the [`private` field proposal for ES6](https://tc39.es/proposal-class-fields/).
```suggestion
  static #getActiveMembers(members) {
```

adamweeks: This doesn't work with our eslint config. I'll add a `@private` to the jsdoc.

src/MembershipsSDKAdapter.test.js
adamweeks: Since we will have different tests for the different destination types, it is easier to group the tests into the blocks for each situation. I will change the describe block so that the test reads a bit better:
`Memberships SDK Adapter getMembersFromDestination() when destination type is ROOM throws an error on subscription`

src/MembershipsSDKAdapter.test.js
adamweeks: when the room memberships functionality is added, the tests will go in this describe block.

src/MembershipsSDKAdapter.js
```diff
@@ -0,0 +1,110 @@
...
+    return Object.values(members)
+      .filter((member) => !member.isInMeeting)
+      .map((member) => ({
+        id: member.id,
+      }));
+  }
+
+  constructor(datasource) {
```
lalli-flores, line 22: Can I suggest an order of methods for a class? Maybe I can open a PR later in the contribution guide to make sure it's standardized along all of our classes.
This is how I would suggest the order:
1. Constructor <- Class constructor always goes first
2. Static methods <- Since they're not part of the instance
3. Interface-implementation methods <- Since component system users will be familiar with those the order should match that of the interface
4. Other helper/private methods

src/MembershipsSDKAdapter.js
lalli-flores: Thinking about this function again, I don't think it should be a method of the class. I think this can work ok being a local module function (that does **_not_** get exported), as:
1. We don't want to expose this as part of the JSON adapter object
2. Only other functions/methods in this module should make use of it

src/MembershipsSDKAdapter.js
```diff
@@ -0,0 +1,110 @@
...
+
+  /**
+   * Gets a shared rxjs subject for the given membership
+   *
+   * @private
+   * @param {string} membershipID  ID of the destination for which to get members
+   * @param {DestinationType} destinationType  type of the membership destination
+   * @returns {BehaviorSubject}
+   */
+  getOrCreateMembershipSubject(membershipID) {
```
lalli-flores, line 36: hahaha very Java-esc name, maybe simplifying it to "get" should be sufficient.
I suggest we can expand on the docstring that it will be created if it doesn't exist
```suggestion
  getMembershipSubject(membershipID) {
```

src/MembershipsSDKAdapter.js
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  constructor(datasource) {
+    super(datasource);
+
+    this.membershipSubjects = {};
+  }
+
+  /**
+   * Gets a shared rxjs subject for the given membership
```
lalli-flores, line 29:
```suggestion
   * Returns an observable subject for the given membership.
   * If observable subject does not exist, it will be created.
```

src/MembershipsSDKAdapter.js
```diff
@@ -0,0 +1,110 @@
...
+
+    const meeting = this.datasource.meetings.getMeetingByType('id', destinationID);
+
+    if (!meeting) {
+      return throwError(new Error(`Meeting ${destinationID} not found.`));
+    }
+
+    const membershipID = `${destinationType}-${destinationID}`;
+
+    if (!this.membershipSubjects[membershipID]) {
```
lalli-flores, line 69: Didn't you create a helper method for this?

src/MembershipsSDKAdapter.js
```diff
@@ -0,0 +1,110 @@
...
+        ID: membershipID,
+        destinationID,
+        destinationType,
+        members: [],
+      });
+    }
+
+    const membershipSubject = this.membershipSubjects[membershipID];
+
+    if (meeting.members && meeting.members.membersCollection && meeting.members.membersCollection.members) {
```
lalli-flores, line 82: Sidenote: we should look into adding conditional chaining 😛

lalli-flores (commented):
Thanks for re-arranging/naming the tests, much clearer 👍🏼

src/MembershipsSDKAdapter.js
adamweeks: Yep, I tried it but we didn't have the babel plugin installed and didn't want to do it as part of this PR

src/MembershipsSDKAdapter.js
adamweeks: The name works fine as is.

src/MembershipsSDKAdapter.js
adamweeks: I did but was not needed. Over optimized.

mwegman (approved):

src/MembershipsSDKAdapter.js
lalli-flores: Then why didn't you remove it?

src/MembershipsSDKAdapter.js
```diff
@@ -0,0 +1,109 @@
...
+    if (!meeting) {
+      return throwError(new Error(`Meeting ${destinationID} not found.`));
+    }
+
+    const membershipID = `${destinationType}-${destinationID}`;
+
+    if (!this.membershipSubjects[membershipID]) {
+      // Behavior subject will keep the last emitted object for new subscribers
+      // https://rxjs.dev/guide/subject#behaviorsubject
+      this.membershipSubjects[membershipID] = new BehaviorSubject({
```
lalli-flores, line 71: You should initialize with the data after you fetched it. No need to emit twice to send empty + initial data

src/MembershipsSDKAdapter.js
adamweeks: https://github.com/webex/sdk-component-adapter/pull/80

I did, but it came back after a rebase apparently

PR #79, opened 2020-09-10
Server improvements

adamweeks:
This adds features to the `npm run serve` functionality that allows the user to input an access token instead of pulling from the .env file (thus making it more useful for automation tests in the future). Also adds membership support but is blocked by #77. 

The browserslist addition to `package.json` is required to be able to run `npm run serve` with the latest `webex` package. I've run a `npm run build` and `npm link` for the package to verify the browserslist addition didn't break anything and it works as expected when used in the widget repo.

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,5 +120,8 @@
       "pre-commit": "npm run test:eslint",
       "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
     }
-  }
+  },
+  "browserslist": [
+    "last 2 Chrome versions"
+  ]
 }
--- a/scripts/index.html
+++ b/scripts/index.html
@@ -9,6 +9,7 @@
         <form id="connector">
           <fieldset>
             <legend>JS SDK Connection</legend>
+            <input id="credentials" placeholder="Access Token" type="password" />
             <button id="connect">Connect</button>
             <button id="disconnect">Disconnect</button>
             <span id="connection-status"> Disconnected ❌ </span>
@@ -29,6 +30,13 @@
             <button id="mute-audio">Mute Audio</button>
             <button id="mute-video">Mute Video</button>
             <button id="share-screen">Screen Share</button>
+          </fieldset>
+        </form>
+        <form id="members">
+          <fieldset>
+            <legend>Members</legend>
+            <button id="get-members">Get Members</button>
+            <textarea id="members-list"></textarea>
           </fieldset>
         </form>
         <div style="display: flex; justify-content: space-between;">
--- a/scripts/puppeteer.server.js
+++ b/scripts/puppeteer.server.js
@@ -1,6 +1,11 @@
-import webexSDKAdapter from './start';
+/* eslint-disable no-console */
+import {DestinationType} from '@webex/component-adapter-interfaces';
+import Webex from 'webex';
+
+import WebexSDKAdapter from '../src/WebexSDKAdapter';
 
 let MEETING_ID = null;
+let webexSDKAdapter;
 
 function handleAudio() {
   webexSDKAdapter.meetingsAdapter.meetingControls['mute-audio'].display(MEETING_ID).subscribe((data) => {
@@ -19,9 +24,7 @@ function handleVideo() {
 }
 
 function handleShare() {
-  webexSDKAdapter.meetingsAdapter.meetingControls['share-screen']
-    .display(MEETING_ID)
-    .subscribe((data) => {
+  webexSDKAdapter.meetingsAdapter.meetingControls['share-screen'].display(MEETING_ID).subscribe((data) => {
     const startShare = document.getElementById('share-screen');
 
     startShare.innerHTML = data.tooltip;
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -120,5 +120,8 @@
       "pre-commit": "npm run test:eslint",
       "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
     }
-  }
+  },
+  "browserslist": [
```
lalli-flores, line 124: Is this necessary because we don't have the latest version of babel? I know we had a warning for `browserlist` when we ran the tests in the other repos, but it went away as I upgraded all of our dev dependencies there.
I haven't had a chance to get to this repo yet

scripts/index.html
```diff
@@ -6,6 +6,15 @@
         <h1 id="page-title">
           Meeting Integration Tests Sample
         </h1>
+        <form id="adapter">
+          <fieldset>
+            <legend>JS SDK Adapter</legend>
+            <input id="credentials" placeholder="Access Token" type="password" />
+            <button id="create">Create</button>
```
lalli-flores, line 13:
```suggestion
            <button id="create-adapter">Create</button>
```

scripts/index.html
```diff
@@ -6,6 +6,15 @@
...
+            <button id="destroy">Destroy</button>
```
lalli-flores, line 14:
```suggestion
            <button id="destroy-adapter">Destroy</button>
```

Just to make it more specific if/when things start to expand

scripts/puppeteer.server.js
```diff
@@ -111,7 +138,25 @@ document.getElementById('actions').addEventListener('click', async (event) => {
...
+  }
+});
+
+document.getElementById('members').addEventListener('click', async (event) => {
+  event.preventDefault();
+
+  try {
+    switch (event.target.id) {
+      case 'get-members':
+        console.log(webexSDKAdapter.membershipsAdapter);
```
lalli-flores, line 151: Was this for testing or are you logging it purposely?

scripts/puppeteer.server.js
```diff
@@ -111,7 +138,25 @@ document.getElementById('actions').addEventListener('click', async (event) => {
...
+        await webexSDKAdapter.membershipsAdapter
```
lalli-flores, line 152: Is the `await` necessary? isn't this an observable?
```suggestion
       webexSDKAdapter.membershipsAdapter
```

scripts/puppeteer.server.js
```diff
@@ -111,7 +138,25 @@ document.getElementById('actions').addEventListener('click', async (event) => {
...
+    console.error('Unable to perform an action:', error);
+  }
+});
+
+document.getElementById('members').addEventListener('click', async (event) => {
+  event.preventDefault();
+
+  try {
+    switch (event.target.id) {
+      case 'get-members':
```
lalli-flores, line 150: Do you foresee any other short-term events being added to `members`? Otherwise we shouldn't use a switch statement here. YAGNI

lalli-flores:
I like the idea of making this something we can use for automation tests 👍🏼

package.json
adamweeks: Yep, otherwise `npm run serve` does not work. That's why I mentioned it directly in the PR description.

scripts/puppeteer.server.js
```diff
@@ -1,6 +1,11 @@
-import webexSDKAdapter from './start';
```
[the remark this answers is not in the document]
adamweeks: The start script is still good for testing adapters that don't have browser components.

scripts/index.html
adamweeks: These are used as nested id's in the implementation so it is the "destroy" element within the "adapter".

scripts/puppeteer.server.js
```diff
@@ -42,12 +44,39 @@ function getMeeting() {
...
+
+  try {
+    switch (event.target.id) {
+      case 'create':
+        {
+          const webex = new Webex({
+            credentials,
+          });
+
+          webexSDKAdapter = new WebexSDKAdapter(webex);
```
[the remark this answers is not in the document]
adamweeks: Not sure what you mean by "registration". There is a `connect` section below this which does the `adapter.connect`.

scripts/puppeteer.server.js
```diff
@@ -42,12 +44,39 @@ function getMeeting() {
...
+          status.innerHTML = 'Created ✅';
+        }
+        break;
+      case 'destroy':
+        webexSDKAdapter = null;
```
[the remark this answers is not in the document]
adamweeks: We do so in the disconnect section.

scripts/index.html
lalli-flores: There is no such thing as "nested IDs" in HTML

scripts/puppeteer.server.js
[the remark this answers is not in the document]
lalli-flores: Then why make it more complicated by having a separate "create"? Can we have one or the other? and one step does both, create + connect.

scripts/puppeteer.server.js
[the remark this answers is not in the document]
lalli-flores: Same as my previous comment, no need to have to separate steps, can you combine the disconnect + destroy?

adamweeks:
@lalli-flores addressed

lalli-flores (approved):
Thanks @adamweeks!

PR #81, opened 2020-09-14
Remote share event support

adamweeks:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-175408

Manually tested with the server and it removed/added remote share in the dom as expected!

### DIFF
--- a/src/MeetingsSDKAdapter.js
+++ b/src/MeetingsSDKAdapter.js
@@ -8,6 +8,8 @@ const EVENT_MEDIA_READY = 'media:ready';
 const EVENT_MEDIA_STOPPED = 'media:stopped';
 const EVENT_LOCAL_SHARE_STOP = 'meeting:stoppedSharingLocal';
 const EVENT_LOCAL_SHARE_START = 'meeting:startedSharingLocal';
+const EVENT_REMOTE_SHARE_START = 'meeting:startedSharingRemote';
+const EVENT_REMOTE_SHARE_STOP = 'meeting:stoppedSharingRemote';
 
 // Adapter Events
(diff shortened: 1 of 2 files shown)


src/MeetingsSDKAdapter.js
```diff
@@ -181,7 +183,15 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
         this.meetings[ID] = {...meeting, localShare: stream};
         break;
       case MEDIA_TYPE_REMOTE_SHARE:
-        this.meetings[ID] = {...meeting, remoteShare: stream};
+        this.meetings[ID] = {...meeting, remoteShareStream: stream};
```
adamweeks, line 186: Since this is the only event we get the with remote share stream, I'm storing it in a property on the meeting object and then using it when we get the stop/start events.

src/MeetingsSDKAdapter.js
```diff
@@ -181,7 +183,15 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
+        break;
+      case EVENT_REMOTE_SHARE_START:
+        // Only activate the remote stream when get get the start notification
+        this.meetings[ID] = {...meeting, remoteShare: meeting.remoteShareStream};
```
adamweeks, line 190: We also have the option to create a new stream every time from the same track. I figured the way it was implemented cause less object creation and memory consumption:

```js
this.meetings[ID] = {...meeting, remoteShare: new MediaStream([meeting.mediaProperties.remoteShare])};
```

mwegman (approved):

PR #83, opened 2020-09-24
feat(package): update webex to 1.96.1

adamweeks:


lalli-flores:
Does it make sense to fix the `webex` version?
Additionally, can we get it removed from the dev dependencies? There is an easy command to install peer dependencies locally for development, `npx npm-install-peers` that we should encourage to use instead.

lalli-flores (changes requested):
Also, not sure if the commit should release a new version of the package? Maybe a patch if at all since it's a peer dependency

adamweeks:
> Does it make sense to fix the `webex` version?
> Additionally, can we get it removed from the dev dependencies? There is an easy command to install peer dependencies locally for development, `npx npm-install-peers` that we should encourage to use instead.

My concern with fixing the version is that if someone wanted to use a newer version of the sdk with the components or widgets, it would complain about a missing peer dependency. As long as the major version matches, we should technically be good feature wise.

lalli-flores:
I'm with not fixing it as long we don't generate a minor version update and we remove `webex` from the dev dependencies

adamweeks:
Not needed as the semver will pick it up and no new features are required.

PR #86, opened 2020-12-02
Linter updates

lalli-flores:
Updates linter to adhere with agreed team standards.


src/testHelper.js
```diff
@@ -73,8 +75,8 @@ export async function removeIntegrationTestUser(user) {
 /**
  * Constructs a Hydra ID for a given UUID.
  *
- * @param {String} UUID Person UUID
- * @returns {string}
+ * @param {string} UUID Person UUID
```
adamweeks, line 78: Please help me find a linter that will correct the capitalization for these! JSDoc site has `string` lowercase and `Object` uppercase and I can never remember which is which: https://jsdoc.app/tags-param.html

taymoork2: Try this? https://github.com/aladdin-add/eslint-plugin/tree/master/packages/autofix

lalli-flores: The `eslint-plugin-jsdoc` actually checks for that. They [have an explanation on when things are what](https://github.com/gajus/eslint-plugin-jsdoc#why-not-capital-case-everything), and the rule is defined as [check-types](https://github.com/gajus/eslint-plugin-jsdoc#eslint-plugin-jsdoc-rules-check-types)

adamweeks (approved):

PR #91, opened 2021-03-11
refactor(MembershipsSdkAdapter): prevent multiple identical emits on …

cipak:
…member updates

This PR aims to simplify the management of the membership observable. If the observable for a certain destination already exists, it is returned without any additional setup (e.g., no new `mebers:update` listener is registered).

TODO: clear the 'cache' and unregister the event listener when there are no more subscribers to a membership observable.

### DIFF
--- a/src/MembershipsSDKAdapter.js
+++ b/src/MembershipsSDKAdapter.js
@@ -17,13 +17,13 @@ import {DestinationType, MembershipsAdapter} from '@webex/component-adapter-inte
  * @returns {Array} List of active users in a meeting
  */
 function getMembers(members) {
-  return Object.values(members)
+  return members ? Object.values(members)
     .filter((member) => member.isUser)
     .map((member) => ({
       id: member.id,
       inMeeting: member.inMeeting,
       muted: member.isAudioMuted,
-    }));
+    })) : [];
 }
 
 /**
@@ -36,7 +36,7 @@ export default class MembershipsSDKAdapter extends MembershipsAdapter {
   constructor(datasource) {
     super(datasource);
 
-    this.membershipSubjects = {};
+    this.membership$ = {}; // cache membership observables based on membership id
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/MembershipsSDKAdapter.js
```diff
@@ -51,42 +51,33 @@ export default class MembershipsSDKAdapter extends MembershipsAdapter {
...
+    }
+
     const meeting = this.datasource.meetings.getMeetingByType('id', destinationID);
 
     if (!meeting) {
       return throwError(new Error(`Meeting ${destinationID} not found.`));
     }
 
-    const membershipID = `${destinationType}-${destinationID}`;
+    const members = meeting.members
```
lalli-flores, line 66: Maybe rename this to make it a little bit clearer that is meant as a boolean?
```suggestion
    const hasMembers = meeting.members
```

src/MembershipsSDKAdapter.js
```diff
@@ -51,42 +51,33 @@ export default class MembershipsSDKAdapter extends MembershipsAdapter {
...
-        members: [],
-      });
-    }
+    // Behavior subject will keep the last emitted object for new subscribers
+    // https://rxjs.dev/guide/subject#behaviorsubject
+    this.membershipSubjects[membershipID] = new BehaviorSubject({
+      ID: membershipID,
+      destinationID,
+      destinationType,
+      members: members ? getActiveMembers(members) : [],
```
lalli-flores, line 76:
```suggestion
      members: hasMembers ? getActiveMembers(members) : [],
```

lalli-flores (commented):
Thanks for refactoring this! Looks clean, just a minor comment

src/MembershipsSDKAdapter.js
cipak: it's actually the member list from the sdk, but "prudently" extracted from the inner data structure of member collection.

If we had the optional chaining operator, we could write a more concise:
```js
const members = meeting.members?.membersCollection?.members;
```

This replaces the `if` on line 75 of the previous version of this file.

If you prefer, I can move the check inside the `getActiveMembers` function, so that it returns an empty array if it receives a nullish `members` argument. As a benefit of this approach, a `members:update` event with an empty `payload.full` property would no longer generate an error:

```js
const updatedMembers = getActiveMembers(payload.full);
```

lalli-flores: I like your suggestion on refactoring `getActiveMembers`, it's cleaner 👍🏼

cipak: thanks!

cipak: @lalli-flores I rebased this PR onto the (fixed) master branch and implemented the null check in `get(Active)Members`

lalli-flores:
isn't this code similar to what is in https://github.com/webex/sdk-component-adapter/pull/94?

cipak:
> isn't this code similar to what is in #94?

I don't know why your comment above can't be normally replied to, so I had to use "quote reply":) Yes, they share some code.

This PR fixes the double event listener registration (and subsequent double observable emits) that was done when the `getMembersFromDestination` function was called a second time with the same destination id.

PR #94 is built on top of this one (not from master). When this one will be merged, only specific diffs from #94 will remain, which are related to making `getMembersFromDestination` emit member lists and not whole membership objects.

lalli-flores (approved):
Beautiful code! Thanks for making the refactoring changes👍🏼

PR #93, opened 2021-03-19
feat(MembershipsSdkAdapter): return members of a room

cricku:
add the ability to return members of a room using `roomID` as an identifier 

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-154240

### DIFF
--- a/src/MembershipsSDKAdapter.js
+++ b/src/MembershipsSDKAdapter.js
@@ -1,5 +1,26 @@
-import {BehaviorSubject, throwError} from 'rxjs';
-import {DestinationType, MembershipsAdapter} from '@webex/component-adapter-interfaces';
+import {
+  BehaviorSubject,
+  concat,
+  defer,
+  fromEvent,
+  merge,
+  throwError,
+} from 'rxjs';
+import {
+  filter,
+  finalize,
+  map,
+  mergeMap,
+  publishReplay,
+  refCount,
+} from 'rxjs/operators';
+import {SDK_EVENT, deconstructHydraId} from '@webex/common';
+import {
+  DestinationType,
+  MembershipsAdapter,
+} from '@webex/component-adapter-interfaces';
+
+const MAX_MEMBERSHIPS = 5000; // maximum participants in a space
 
 // TODO: Figure out how to import JS Doc definitions and remove duplication.
 /**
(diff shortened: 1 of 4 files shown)


src/MembershipsSDKAdapter.js
```diff
@@ -38,6 +44,32 @@ export default class MembershipsSDKAdapter extends MembershipsAdapter {
...
 
+  /**
+   * Gets members from a room
+   *
+   * @private
+   * @param {string} roomID ID of the room
+   * @returns {Array} List of members in a room
+   */
+  getRoomMembers(roomID) {
+    const roomMembers$ = new BehaviorSubject([]);
```
cipak, line 55: just `members$` should be fine in this context

src/MembershipsSDKAdapter.js
```diff
@@ -38,6 +44,32 @@ export default class MembershipsSDKAdapter extends MembershipsAdapter {
...
+
+    from(this.datasource.memberships.list({
```
lalli-flores, line 57: For promises we prefer to use the [`defer` operator](http://reactivex.io/documentation/operators/defer.html). This is because when you use `from` with a promise, the promise gets executed as soon as the code gets loaded and then the promise resolution gets cached, so the promise doesn't get executed again for the lifetime of the execution. When using `defer` the promise gets executed every time theres a new subscriber, as the promise resolution is deferred until it is called/subscribed.
In this case, if there happens to be an update the memberships of the room since the code started running, the observable using `from` won't pickup on it (since the promise resolved long back) and a new subscriber after the membership changed won't get the benefit of having this new information.

lalli-flores (commented):
Looks good so far. Don't forget to utilize [`memberships.listen()`](https://webex.github.io/webex-js-sdk/api/#membershipslisten) to also emit whenever a user is added to the room (`created`) and when a user gets removed (`deleted`)

src/MembershipsSDKAdapter.js
```diff
@@ -38,6 +44,32 @@ export default class MembershipsSDKAdapter extends MembershipsAdapter {
     this.membershipSubjects = {};
   }
 
+  /**
+   * Gets members from a room
```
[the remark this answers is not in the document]
cipak: @lalli-flores As the listening can only be done on all room memberships (not just the "current" room), we'll have to filter the received events based on roomId. We'll be basically looking for `created` and `deleted` events (maybe `updated` events as well). Also, for simplicity and to preserve the backend order, we were thinking of just reloading the members of a room after such events, instead of trying to amend the member list by pushing the new member at the end of the previously loaded list or filtering out the deleted one. We'd be interesting in hearing your thoughts on this.

src/MembershipsSDKAdapter.js
cipak: There is that caching mechanism in place (`membershipSubjects`) that updates the member lists based on events and creates a single `BehaviorSubject` observable for each destination (room/meeting). So, in theory, a single backend call for loading the members would be needed, the rest being done via sdk events. However, because we are planning to reload the room members after each added/deleted event, it makes sense to be able to re-run this backend call. In any case, we'll keep this in mind and prefer defer over from.

lalli-flores: @cipak, you're correct, because we're caching we should technically only run the initial query once. Using `defer` has the side benefit of matching the mental execution flow when debugging. When we were first creating some of the other adapters and try to debug, we were really confused as to why the promise was running "off cycle", but then we realize why (what I explained above).

We should be careful to clean up the cache when the subject has no subscribers, as the subject would have completed at that point.

src/MembershipsSDKAdapter.js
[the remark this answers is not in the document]
lalli-flores: > @lalli-flores As the listening can only be done on all room memberships (not just the "current" room), we'll have to filter the received events based on roomId. We'll be basically looking for `created` and `deleted` events (maybe `updated` events as well). 
Yes, that's the way to do it. No need to look at `updated`, that is for the read receipts mostly, which we are not implementing at this point.

> Also, for simplicity and to preserve the backend order, we were thinking of just reloading the members of a room after such events, instead of trying to amend the member list by pushing the new member at the end of the previously loaded list or filtering out the deleted one. We'd be interesting in hearing your thoughts on this.

It's ok to send a brand new array from the SDK adapter, not completely sure I like the idea of reloading the whole list in UI each time though. That's a UI issue though, which we should be addressed separately.
As for what the list should look like, please remember that the final list should have the current user at the top, following an alphabetical list of the rest of the members.
Please, [see the Jira for full requirements on this feature](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-154240)

src/MembershipsSDKAdapter.js
```diff
@@ -37,6 +56,125 @@ export default class MembershipsSDKAdapter extends MembershipsAdapter {
...
+  /**
+   * Tells the SDK to start listening to memberships events and tracks the amount of calls.
+   *
+   * Note: Since the SDK listens to ALL memberships events, this function only
+   * calls the SDK's `memberships.listen` function on the first membership to listen.
+   * Repeated calls to `memberships.listen` are not needed afterwards.
+   *
+   * @private
+   */
+  startListeningToMembershipsUpdates() {
```
lalli-flores, line 71: Do you think it would be a good idea to follow the MeetingsAdapter pattern of having a `connect` and `disconnect` method than then gets called with [`WebexSDKAdapter.connect()` as part of the global setup](https://github.com/webex/sdk-component-adapter/blob/master/src/WebexSDKAdapter.js#L26) instead of keeping a counter?

lalli-flores (commented):
Minor comments and questions but otherwise looks really good!

src/MembershipsSDKAdapter.js
```diff
@@ -37,6 +56,125 @@ export default class MembershipsSDKAdapter extends MembershipsAdapter {
...
+      id: deconstructHydraId(membership.personId).id,
+      orgID: deconstructHydraId(membership.personOrgId).id,
+      muted: null,
+      sharing: null,
+      inMeeting: null,
+    });
+
+    const members$ = defer(() => this.datasource.memberships.list({
+      roomId: roomID,
+      max: 1000,
```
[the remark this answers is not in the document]
cricku: I will move that value in a constant. I knew that `memberships.list()` method haves a default limit of `100 ` so we used that large number in order to get all the memberships in only one request. Is there a maximum number of members that can be part of a room?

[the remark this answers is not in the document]
lalli-flores: According to the [official docs](https://help.webex.com/en-us/n8vw82eb/Webex-Capacities) max number of people in a space is 5000.
Does changing 100 -> 1000 have performance implications or are you using a paginated approach?

src/MembershipsSDKAdapter.js
cricku: Using the same approach as for the `MeetingsAdapter` the WebSocket will start emitting events right after we call `connect()` on the adapter but for example, we don't have any subscribers to the `getMembersFromDestination()`. I think this approach will just keep the WebSocket Network busy with events that maybe we don't want at that point.

While using a counter in `MembershipsAdapter` we will start listening to events only when someone subscribes to `getMembersFromDestination()`.

lalli-flores (approved):

PR #94, opened 2021-03-19
refactor(MembershipsSdkAdapter): emit member lists from getMembersFromDestination

cipak:
getmembersFromDestination should return an observable that emits member lists (instead of membership objects).

Contains changes from PR https://github.com/webex/sdk-component-adapter/pull/91, which should be merged first.

### DIFF
--- a/src/MembershipsSDKAdapter.js
+++ b/src/MembershipsSDKAdapter.js
@@ -3,9 +3,9 @@ import {DestinationType, MembershipsAdapter} from '@webex/component-adapter-inte
 
 // TODO: Figure out how to import JS Doc definitions and remove duplication.
 /**
- * A relationship between a destination (e.g. room, meeting) in Webex and people.
+ * A Member object that is part of a membership
  *
- * @external Membership
+ * @external Member
  * @see {@link https://github.com/webex/component-adapter-interfaces/blob/master/src/MembershipsAdapter.js#L6}
  */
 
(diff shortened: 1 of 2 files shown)


src/MembershipsSDKAdapter.js
```diff
@@ -40,66 +40,45 @@ export default class MembershipsSDKAdapter extends MembershipsAdapter {
...
     if (!meeting) {
       return throwError(new Error(`Meeting ${destinationID} not found.`));
     }
 
-    const membershipID = `${destinationType}-${destinationID}`;
+    const members = meeting.members
+      && meeting.members.membersCollection
+      && meeting.members.membersCollection.members;
 
-    if (!this.membershipSubjects[membershipID]) {
```
lalli-flores, line 64: I remember now why we had it like this. For this code base, we want to keep a single exit control. In my experience, code is easier to debug if you can always expect the code to culminate in the same place.
Would you mind updating taking this into account?
So code would be something like:
```js
// Remember we're using the $ convention for observables
// Could probably update this.membershipSubjects => this.membersSubjects
let members$ = this.membershipSubjects[membershipID];

if (destinationType !== DestinationType.MEETING) {
 members$ = throwError(..)
}

if(!meeting) {
 members$ = throwError(..)
}

if(!members$) {
members$ = new BehaviorSubject(...)
...
}

return members$

Comment (cipak) [reply src/MembershipsSDKAdapter.js:64]: I personally prefer early returns as they "free my mind" of exception cases while I read the code. For example, I can mentally dismiss the case of destination type not being "room" when I see that return and focus just on the remaining case. Also, this helps reduce indentation, as the "main case" does not need to be included in an `if` or `else` block.

However, I also think that convention between team members is more important than personal preferences (which are different most often than not), so I will implement this with a single `return`.

Comment (cipak) [reply src/MembershipsSDKAdapter.js:64]: I did the refactoring, but I find the if-else structure a little hard to read.

Compare the early return variant https://github.com/webex/sdk-component-adapter/blob/6be957381f989fcbbd3fdc75d47412137927a889/src/MembershipsSDKAdapter.js#L51 with the single return https://github.com/webex/sdk-component-adapter/blob/f7c15b13d9e0dbaa961a324d2c5d5c58ab7781cd/src/MembershipsSDKAdapter.js#L51

 I also tried flattening the `if-else` structure but the conditions look a little repetitive with the omnipresent `!membership$`:
```js
  getMembersFromDestination(destinationID, destinationType) {
    const membershipID = `${destinationType}-${destinationID}`;
    let membership$ = this.membership$[membershipID];

    if (!membership$ && destinationType !== DestinationType.MEETING) {
      membership$ = throwError(new Error(`getMembersFromDestination for ${destinationType} is not currently supported.`));
    }

    let meeting;

    if (!membership$) {
      meeting = this.datasource.meetings.getMeetingByType('id', destinationID);
    }

    if (!membership$ && !meeting) {
      membership$ = throwError(new Error(`Meeting ${destinationID} not found.`));
    }

    if (!membership$) {
      const members = meeting.members
        && meeting.members.membersCollection
        && meeting.members.membersCollection.members;

      // Behavior subject will keep the last emitted object for new subscribers
      // https://rxjs.dev/guide/subject#behaviorsubject
      membership$ = new BehaviorSubject({
        ID: membershipID,
        destinationID,
        destinationType,
        members: getMembers(members),
      });

      this.membership$[membershipID] = membership$; // save for future calls

      // Emit on membership updates
      meeting.members.on('members:update', (payload) => {
        if (payload && payload.full) {
          const updatedMembers = getMembers(payload.full);

          membership$.next({
            ID: membershipID,
            destinationID,
            destinationType,
            members: updatedMembers,
          });
        }
      });
    }

    return membership$;
}
```

If you have an idea on how could I better format this code, please let me know.

lalli-flores (approved):
Looks good! Thanks for making code clearer!

PR #90, opened 2021-03-11
feat(MembershipsSDKAdapter): return membership status and all members

maloun96:

PR #95, opened 2021-03-21
fix(MembershipsSdkAdapter): fix test

maloun96:
Fix tests and bugs that were introduced in the master branch in this PR
https://github.com/webex/sdk-component-adapter/pull/90
@lalli-flores


lalli-flores (approved):
Thanks for the fix!

Seems like checks are still not running for you. We need to fix this, it has become a problem as the team can't confidently merge code changes that we don't know has pass automation.

PR #96, opened 2021-03-24
fix: fetch room data with globalId

mwegman:


adamweeks (approved):

src/MeetingsSDKAdapter.js
```diff
@@ -291,7 +291,7 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
 
       meetingTitle = `${displayName}'s Personal Room`;
     } else if (type === HYDRA_ID_TYPE_ROOM) {
-      const {title} = await this.datasource.rooms.get(id);
+      const {title} = await this.datasource.rooms.get(destination);
```
lalli-flores, line 294: can you add a comment here?
```suggestion
      // To find a room cluster + UUID are necessary, both encoded in the destination
      const {title} = await this.datasource.rooms.get(destination);
```

PR #97, opened 2021-03-25
fix(memberships): update Webex SDK

mwegman:
- 'member' object changed a property name.


lalli-flores (approved):
Thanks for finding and fixing it quickly!

lalli-flores:
@mwegman [mock SDK](https://github.com/webex/sdk-component-adapter/blob/master/src/mockSdk.js) needs to be updated to match the changes in the SDK.

PR #98, opened 2021-03-29
feat(MembershipsSdkAdapter): return the sharing status of each member

lungancatalin:
Update SDK Adapter to send a sharing status for a member

### DIFF
--- a/src/MembershipsSDKAdapter.test.js
+++ b/src/MembershipsSDKAdapter.test.js
@@ -28,10 +28,12 @@ describe('Memberships SDK Adapter', () => {
               {
                 id: 'id',
                 muted: false,
+                sharing: false,
               },
               {
                 id: 'mutedPerson',
                 muted: true,
+                sharing: true,
               },
               {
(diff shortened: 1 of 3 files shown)


src/MembershipsSDKAdapter.test.js
```diff
@@ -32,10 +32,12 @@ describe('Memberships SDK Adapter', () => {
                 {
                   id: 'id',
                   muted: false,
+                  sharing: false,
                 },
                 {
                   id: 'mutedPerson',
                   muted: true,
+                  sharing: true,
```
lalli-flores, line 40: It's funny that the muted person is sharing 😂

lalli-flores (commented):
Seems like you need to update your branch. There are conflicts with master

src/MembershipsSDKAdapter.test.js
lungancatalin: Yes, it's funny, but it could be a real case :))

lalli-flores (approved):

PR #101, opened 2021-04-01
feat(MembershipsSdkAdapter): add orgID field to meeting members

maloun96:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-220419

I changed mockSdk to match with real data from webex.

The real data:
image

The previous mock data
image

### DIFF
--- a/src/MembershipsSDKAdapter.js
+++ b/src/MembershipsSDKAdapter.js
@@ -21,6 +21,7 @@ function getMembers(members) {
     .filter((member) => member.isUser)
     .map((member) => ({
       id: member.id,
+      orgID: member.participant && member.participant.person && member.participant.person.orgId,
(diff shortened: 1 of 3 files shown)


src/MembershipsSDKAdapter.js
```diff
@@ -24,6 +24,7 @@ function getMembers(members) {
       inMeeting: member.isInMeeting,
       muted: member.isAudioMuted,
       sharing: member.isContentSharing,
+      orgID: member.orgID,
```
cipak, line 27: move this up just after `id` please, so we keep the same order as in the interface

cipak:
I would suggest this commit message "add orgID field to meeting members"

lalli-flores (approved):

PR #102, opened 2021-04-01
feat(OrganizationsSdkAdaper): add organizations adapter

cricku:
Implement organizations adapter

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-220451

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
   },
   "dependencies": {
     "@webex/common": "^1.105.0",
-    "@webex/component-adapter-interfaces": "^1.13.0"
+    "@webex/component-adapter-interfaces": "^1.16.0"
   },
   "devDependencies": {
     "@babel/core": "^7.12.3",
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
   },
   "dependencies": {
     "@webex/common": "^1.105.0",
-    "@webex/component-adapter-interfaces": "^1.11.0"
+    "@webex/component-adapter-interfaces": "^1.13.0"
   },
   "devDependencies": {
     "@babel/core": "^7.12.3",
--- a/src/OrganizationsSDKAdapter.js
+++ b/src/OrganizationsSDKAdapter.js
@@ -1,0 +1,69 @@
+import {ReplaySubject, defer} from 'rxjs';
+import {map} from 'rxjs/operators';
+import {OrganizationsAdapter} from '@webex/component-adapter-interfaces';
+
+// TODO: Figure out how to import JS Doc definitions and remove duplication.
+/**
+ * A set of people in Webex.
+ *
+ * @external Organization
+ * @see {@link https://github.com/webex/component-adapter-interfaces/blob/master/src/OrganizationsAdapter.js#L6}
+ */
+
+/**
+ * The `OrganizationsSDKAdapter` is an implementation of the `OrganizationsAdapter` interface.
+ * This adapter utilizes the Webex JS SDK to fetch data about an organization.
+ *
+ * @implements {OrganizationsAdapter}
+ */
+export default class OrganizationsSDKAdapter extends OrganizationsAdapter {
+  constructor(datasource) {
+    super(datasource);
+
+    this.organizationObservables = {};
+  }
+
+  /**
+   * Returns a promise to organization data from Webex.
+   *
+   * @param {string} orgID ID of the organization for which to fetch data
+   * @returns {Promise.<Organization>} Information about the organization of the given ID
+   *
+   * @private
+   */
+  fetchOrganization(orgID) {
+    return this.datasource.request({
+      service: 'hydra',
+      resource: `organizations/${orgID}`,
+    }).then((res) => res.body);
+  }
+
+  /**
+   * Returns an observable that emits organization data of the given ID.
+   *
+   * @param {string} ID ID of organization to get
+   * @returns {external:Observable.<Organization>} Observable stream that emits organization data
+   */
+  getOrg(ID) {
+    if (!(ID in this.organizationObservables)) {
+      // use ReplaySubject cause we don't need to set an initial value
+      this.organizationObservables[ID] = new ReplaySubject(1);
+
+      defer(() => this.fetchOrganization(ID)).pipe(
+        map((response) => ({
+          ID: response.id,
+          name: response.displayName,
+        })),
+      ).subscribe(
+        (organization) => {
+          this.organizationObservables[ID].next(organization);
+        },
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -144,4 +144,4 @@
       ]
     ]
   }
-}
+}
```
cipak, line 147: do we have a convention on end-of-file newlines?

src/OrganizationsSDKAdapter.test.js
```diff
@@ -0,0 +1,59 @@
+import {isObservable} from 'rxjs';
+
+import OrganizationsSDKAdaper from './OrganizationsSDKAdapter';
```
cipak, line 3: adapter:) also on line 8

package.json
```diff
@@ -33,7 +33,7 @@
   },
   "dependencies": {
     "@webex/common": "^1.105.0",
-    "@webex/component-adapter-interfaces": "^1.11.0"
+    "@webex/component-adapter-interfaces": "^1.13.0"
```
lalli-flores, line 36: Things move fast around here! I think latest is now `1.16.0`. Let's use latest so we don't have to update again later. Also, please use a separate `fix` commit for any changes related to dependencies

src/OrganizationsSDKAdapter.js
```diff
@@ -0,0 +1,70 @@
+import {ReplaySubject, defer} from 'rxjs';
+import {map} from 'rxjs/operators';
+import {deconstructHydraId} from '@webex/common';
+import {OrganizationsAdapter} from '@webex/component-adapter-interfaces';
+
+// TODO: Figure out how to import JS Doc definitions and remove duplication.
+/**
+ * A Member object that is part of a membership
```
lalli-flores, line 8:
```suggestion
 * A set of people in Webex.
```

src/OrganizationsSDKAdapter.js
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * @external Organization
+ * @see {@link https://github.com/webex/component-adapter-interfaces/blob/master/src/OrganizationsAdapter.js#L6}
+ */
+
+/**
+ * The `OrganizationsSDKAdapter` is an implementation of the `OrganizationsAdapter` interface.
+ * This adapter utilizes the Webex JS SDK to fetch data about a organization.
```
lalli-flores, line 16:
```suggestion
 * This adapter utilizes the Webex JS SDK to fetch data about an organization.
```

src/OrganizationsSDKAdapter.js
```diff
@@ -0,0 +1,70 @@
...
+    super(datasource);
+
+    this.organizationObservables = {};
+  }
+
+  /**
+   * Fetches the organization data and returns in the shape required by adapter.
+   *
+   * @param {string} orgID ID of the organization for which to fetch data
+   * @returns {Promise<Organization>} Information about the organization of the given ID
```
lalli-flores, line 31:
```suggestion
   * @returns {Promise.<Organization>} Information about the organization of the given ID
```

src/OrganizationsSDKAdapter.js
```diff
@@ -0,0 +1,70 @@
...
+ */
+export default class OrganizationsSDKAdapter extends OrganizationsAdapter {
+  constructor(datasource) {
+    super(datasource);
+
+    this.organizationObservables = {};
+  }
+
+  /**
+   * Fetches the organization data and returns in the shape required by adapter.
```
lalli-flores, line 28: This method doesn't seem to be doing any shaping 😉 
```suggestion
   * Returns a promise to organization data from Webex.
```

src/OrganizationsSDKAdapter.js
```diff
@@ -0,0 +1,70 @@
...
+   * @returns {external:Observable.<Organization>} Observable stream that emits organization data
+   */
+  getOrg(ID) {
+    if (!(ID in this.organizationObservables)) {
+      // use ReplaySubject cause we don't need to set an initial value
+      this.organizationObservables[ID] = new ReplaySubject(1);
+
+      defer(() => this.fetchOrganization(ID)).pipe(
+        map((response) => ({
+          ID: deconstructHydraId(response.id).id,
```
lalli-flores, line 55: Why are you deconstructing the ID? You should send back the "Hydra ID"

src/OrganizationsSDKAdapter.js
```diff
@@ -0,0 +1,70 @@
...
+          name: response.displayName,
+        })),
+      ).subscribe(
+        (organization) => {
+          this.organizationObservables[ID].next(organization);
+        },
+        () => {
+          this.organizationObservables[ID].error(new Error(`Could't find organization with ID ${ID}`));
```
lalli-flores, line 63:
```suggestion
          this.organizationObservables[ID].error(new Error(`Could't find organization with ID "${ID}"`));
```

src/OrganizationsSDKAdapter.test.js
```diff
@@ -0,0 +1,59 @@
...
+            ID: 'organizationID',
+            name: 'Cisco Systems, Inc.',
+          });
+          done();
+        },
+      );
+    });
+  });
+
+  describe('getorg() errors', () => {
```
lalli-flores, line 46: No need to add this extra describe, unless there was some special setup, but that doesn't seem to be the case

package.json
cricku: @lalli-flores do you mean just updating the version to `1.16.0` or using the `latest` tag  e.g: `"@webex/component-adapter-interfaces": "latest"` ?

src/OrganizationsSDKAdapter.js
lalli-flores: Right, the "hydra IDs" are the public API IDs, which are an encoded version of the backend UUIDs + metadata. When interacting with some internal services we need to deconstruct it but not when exposing it outwards.

lalli-flores (commented):
Code looks good! But seems like using `latest` as the version does not work. Maybe try setting it to `^1.16.0`?

package.json
```diff
@@ -33,7 +33,7 @@
   },
   "dependencies": {
     "@webex/common": "^1.105.0",
-    "@webex/component-adapter-interfaces": "^1.11.0"
+    "@webex/component-adapter-interfaces": "latest"
```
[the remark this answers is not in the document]
cricku: locally it's working fine but when the checks are run in Circle CI seems that using the tag `latest` on a package doesn't install that package and that's why the linter fails

lalli-flores (approved):

PR #112, opened 2021-04-12
refactor(MembershipsSdkAdapter): sort the members in a room

CernaianuMihai:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-223396

### DIFF
--- a/src/MembershipsSDKAdapter.js
+++ b/src/MembershipsSDKAdapter.js
@@ -1,7 +1,9 @@
 import {
   BehaviorSubject,
+  combineLatest,
   concat,
   defer,
+  from,
   fromEvent,
   merge,
   throwError,
@@ -45,6 +47,31 @@ function sortMeetingMembers(members) {
     : !member1.name ? +1 // empty names come last
     : !member2.name ? -1
     : member1.name.localeCompare(member2.name))); // alphabetical order
+    /* eslint-enable no-nested-ternary, indent */
+}
+
+/**
+ * A Webex user.
+ *
+ * @external Person
+ * @see {@link https://github.com/webex/component-adapter-interfaces/blob/master/src/PeopleAdapter.js#L6}
+ */
+
+/**
+ * Sort a memberships list alphabetically, with the current user first
+ *
+ * @param {Array} memberships List of sdk memberships
+ * @param {string} myID Id of the current user
+ * @returns {Array} Sorted list of sdk memberships
+ */
+function sortRoomMembers(memberships, myID) {
+  return memberships.sort((m1, m2) => (
+    /* eslint-disable no-nested-ternary, indent */
+    m1.personId === myID ? -1 // current user comes first
+    : m2.personId === myID ? 1
+    : !m2.personDisplayName ? -1 // empty names come last
+    : !m1.personDisplayName ? 1
+    : m1.personDisplayName.localeCompare(m2.personDisplayName))); // alphabetical order
     /* eslint-enable no-nested-ternary, indent */
 }
 
@@ -141,11 +168,17 @@ export default class MembershipsSDKAdapter extends MembershipsAdapter {
       guest: null,
     });
 
-    const members$ = defer(() => this.datasource.memberships.list({
+    const me$ = from(this.datasource.people.get('me'));
+
+    const memberships$ = defer(() => this.datasource.memberships.list({
       roomId: roomID,
       max: MAX_MEMBERSHIPS,
     })).pipe(
-      map((page) => page.items.map(membershipToMember)),
+      map((page) => page.items),
+    );
+
+    const members$ = combineLatest([me$, memberships$]).pipe(
+      map(([me, memberships]) => sortRoomMembers(memberships, me.id).map(membershipToMember)),
     );
 
     const createdEvent$ = fromEvent(
--- a/__mocks__/@webex/common.js
+++ b/__mocks__/@webex/common.js
@@ -1,8 +1,8 @@
 export const deconstructHydraId = (ID) => {
   if (ID.includes('PEOPLE')) return {type: 'PEOPLE'};
   if (ID.includes('ROOM')) return {type: 'ROOM'};
-  if (ID.includes('personID')) return {id: ID};
-  if (ID.includes('organizationID')) return {id: ID};
+  if (ID.includes('personID')) return {type: 'PEOPLE', id: ID, cluster: 'us'};
+  if (ID.includes('organizationID')) return {type: 'ORGANIZATION', id: ID, cluster: 'us'};
 
   return {};
 };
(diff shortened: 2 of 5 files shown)


src/MembershipsSDKAdapter.js
```diff
@@ -117,11 +136,17 @@ export default class MembershipsSDKAdapter extends MembershipsAdapter {
...
+    const currentUser$ = new PeopleSDKAdapter(this.datasource).getMe();
+
     const members$ = defer(() => this.datasource.memberships.list({
       roomId: roomID,
       max: MAX_MEMBERSHIPS,
-    })).pipe(
-      map((page) => page.items.map(membershipToMember)),
+    }));
+
+    const orderedMembers$ = currentUser$.pipe(
```
cipak, line 146: try using [combineLatest](https://rxjs-dev.firebaseapp.com/api/index/function/combineLatest)

src/MembershipsSDKAdapter.js
```diff
@@ -117,11 +136,17 @@ export default class MembershipsSDKAdapter extends MembershipsAdapter {
       inMeeting: null,
     });
 
+    const currentUser$ = new PeopleSDKAdapter(this.datasource).getMe();
```
cipak, line 139: I don't like the fact that a new instance of people's adapter is created. We should be able to use the same one. However, it's a "sibling" adapter, I'm not sure if there's currently any way to access it. @lalli-flores

src/MembershipsSDKAdapter.js
CernaianuMihai: Apparently combineLatest is deprecated and it recommends me to use pipe to map instead
deprecated_combineLatest

cipak: I think the `resultSelector` parameter is deprecated, not the operator itself. Maybe you called the operator like this `combineLatest($a, $b)` instead of `combineLatest([$a, $b])`?

src/MembershipsSDKAdapter.js
cipak: I'm thinking of adding a "link" to the `WebexSDKAdapter` instance from each specific adapter instance. This could be done by sending an extra parameter to each adapter constructor:
```js
this.peopleAdapter = new PeopleSDKAdapter(sdk, this);
```
https://github.com/webex/sdk-component-adapter/blob/master/src/WebexSDKAdapter.js#L16

Having this, we could easily access sibling adapters:
```js
const me$ = this.parent.peopleAdapter.getMe(); // or this.main, this.adapter, this.adapters, this.siblingAdapters, this.siblings, ...
```

src/MembershipsSDKAdapter.js
CernaianuMihai: Yes, i just tried calling the operator like this combineLatest([$a, $b]), but it's still deprecated
deprecated_combine

src/MembershipsSDKAdapter.js
```diff
@@ -119,11 +146,17 @@ export default class MembershipsSDKAdapter extends MembershipsAdapter {
...
+
+    const memberships$ = defer(() => this.datasource.memberships.list({
       roomId: roomID,
       max: MAX_MEMBERSHIPS,
     })).pipe(
-      map((page) => page.items.map(membershipToMember)),
+      map((page) => page.items),
+    );
+
+    const orderedMembers$ = combineLatest([me$, memberships$]).pipe(
```
cipak, line 158: you can now rename `orderedMembers$` to `members$` , to revert some of the changes below.

src/MembershipsSDKAdapter.js
```diff
@@ -119,11 +146,17 @@ export default class MembershipsSDKAdapter extends MembershipsAdapter {
       host: null,
     });
 
-    const members$ = defer(() => this.datasource.memberships.list({
+    const me$ = from(this.datasource.people.get('me'));
```
cipak, line 149: `from()` should be better than `defer()` in this case, because we don't want the `get(me)` call to be performed on each membership event.

cipak (approved):

lalli-flores (commented):
The code looks good, but tests need to be updated properly.
Update test description to `emits a sorted member list on subscription`.
You're also missing tests for the `sortRoomMembers` function

lalli-flores:
> You're also missing tests for the sortRoomMembers function

CernaianuMihai:
I think the test for `getMembersFromDestionation` is enough because `sortRoomMembers` is called inside of that test and it's an internal function.

lalli-flores:
I disagree. I understand that by virtue of using `getMembersFromDestionation`, `sortRoomMembers` is also called but that doesn't make it "tested".

Adding tests for `sortRoomMembers` allows us to document code by explicitly stating what scenarios we expect `sortRoomMembers` to handle. This is useful for new developers that are trying to get familiar with the code base, and for contributors that haven't looked at the code in a while.
This is also helpful for debugging, where output from `getMembersFromDestionation` might not be showing up as expected, and by having separate tests for `sortRoomMembers` we can quickly assess that is not because an issue there.

In terms of testing strategy, helpers like `sortRoomMembers` should be tested with mock data (which doesn't have to be associated with the overall SDK mock), while for public methods we move towards integration-type tests. I would like to start aligning towards this goal.

cipak:
While I can see the benefit of new developers looking at unit tests to get a better understanding of the existing code base, I feel that the disadvantages of unit testing non-public methods outweigh that benefit:
- more difficult "internal guts" refactoring (unit tests need to be updated as well)
- after such an "internal guts" refactoring, the unit tests no longer tell us we did a good job (they fail anyway)
- test duplication, the same or similar scenarios need to be applied at public level but also private level functions
- we are testing the internal implementation instead of just the public contract (api) of a module
- devs will be tempted to avoid creating a new function. In our case, we can easily inline the sorting in the `getRoomMembers` function to avoid the extra tests, which leads to less clear code. (Btw, `getRoomMembers` does not have unit tests, and neither do `get(Active)Members` and all the other non-public functions in this module)

CernaianuMihai:
We will also have to export the `sortRoomMembers` function only for testing it.

CernaianuMihai:
@lalli-flores Anything else that need to be done in this PR or can it be merged?

lalli-flores:
If you don't think it's a good idea to test `sortRoomMembers`, that's ok, but you at least have to make sure that the scenarios are covered by the public API. I don't see a test for showing when there is the "current" user or when there persons with no display name. There should be 3 scenarios tested total.

CernaianuMihai:
@lalli-flores I added a person with empty name and a person that matches with the current user in mockData and I updated the 'emits a sorted member list on subscription' test to test those two as well.

lalli-flores (approved):
Thanks for updating the tests @CernaianuMihai

PR #113, opened 2021-04-19
fix(MeetingsSdkAdapter): prevent JavaScript errors when access to camera/microphone is denied

karinasigartau0798:
When a user disables camera or denies access to it, a javascript error is displayed in the console and the meeting does not start. This PR solves this bug.

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-229325

### DIFF
--- a/src/MeetingsSDKAdapter.js
+++ b/src/MeetingsSDKAdapter.js
@@ -390,14 +390,15 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
       const localStream = new MediaStream();
 
       const localAudio = this.meetings[ID].localAudio || this.meetings[ID].disabledLocalAudio;
-      const audioTracks = localAudio && localAudio.getTracks();
-
-      audioTracks.forEach((track) => localStream.addTrack(track));
-
       const localVideo = this.meetings[ID].localVideo || this.meetings[ID].disabledLocalVideo;
-      const videoTracks = localVideo && localVideo.getTracks();
 
-      videoTracks.forEach((track) => localStream.addTrack(track));
+      if (localAudio) {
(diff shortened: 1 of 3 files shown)


cipak (approved):

src/MeetingsSDKAdapter.js
```diff
@@ -389,14 +389,15 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
       const localAudio = this.meetings[ID].localAudio || this.meetings[ID].disabledLocalAudio;
-      const audioTracks = localAudio && localAudio.getTracks();
-
-      audioTracks.forEach((track) => localStream.addTrack(track));
-
       const localVideo = this.meetings[ID].localVideo || this.meetings[ID].disabledLocalVideo;
-      const videoTracks = localVideo && localVideo.getTracks();
 
-      videoTracks.forEach((track) => localStream.addTrack(track));
+      if (localAudio) {
```
lalli-flores, line 394: You should also include the check for `getTracks()`, as many browsers out there don't have an implementation for `getTracks()`. Same idea applies to video.

```suggestion
      if (localAudio && localVideo.getTracks()) {
```

cipak: maybe a more robust approach would be to create a polyfill/stub for `getTracks()`, so that we don't have to check explicitly each time we use it (I saw there are other places where `getTracks()` is called without testing if it exists or not):

```js
if (!MediaStream.prototype.getTracks) {
  MediaStream.prototype.getTracks = () => [];
}
```

this would mean that we can always call `getTracks` on a MediaStream, and we can safely iterate the result (even if it's an empty array).

lalli-flores (approved):

PR #116, opened 2021-04-28
build(workflow): update dependency on event

maloun96:


.github/workflows/update-dependencies.yml
```diff
@@ -0,0 +1,20 @@
+name: Update dependencies
+
+on: [repository_dispatch]
+
+jobs:
+  update-webex-component-adapter-interfaces:
+    runs-on: ubuntu-latest
+    if: github.event.action == 'update-webex-component-adapter-interfaces'
```
lalli-flores, line 8: Not sure I understand the purpose of this check?

.github/workflows/update-dependencies.yml
```diff
@@ -0,0 +1,20 @@
...
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2-beta
+      - run: |
+          npm install @webex/component-adapter-interfaces --save
+          if [ -n "$(git status --porcelain)" ]; then
```
lalli-flores, line 14: [This is how we should commit back to the repo](https://github.com/actions/checkout#push-a-commit-using-the-built-in-token)

.github/workflows/update-dependencies.yml
```diff
@@ -0,0 +1,20 @@
...
+            git config user.name "GitHub Actions Bot"
+            git config user.email "<>"
+            git add .
+            git commit -m 'fix(@webex/component-adapter-interfaces): update to new version'
```
lalli-flores, line 18:
```suggestion
            git commit -m 'fix(package): update interfaces dependency'
```

lalli-flores (commented):
we're on the right track; left some feedback. 
Please also rename workflow file `adapter-interfaces-update`

PR #118, opened 2021-05-07
fix(MembershipSdkAdapter): change member id to take the value from participant

patricia0817:
Update member id from member object to retrieve the id from participant person id

### DIFF
--- a/src/MembershipsSDKAdapter.js
+++ b/src/MembershipsSDKAdapter.js
@@ -16,11 +16,12 @@ import {
   publishReplay,
   refCount,
 } from 'rxjs/operators';
-import {SDK_EVENT, deconstructHydraId} from '@webex/common';
+import {SDK_EVENT, constructHydraId, deconstructHydraId} from '@webex/common';
 import {
   DestinationType,
   MembershipsAdapter,
 } from '@webex/component-adapter-interfaces';
+import {hydraTypes} from '@webex/common/dist/constants';
 
 // max parameter value must be greater than 0 and less than or equal to 1000
 const MAX_MEMBERSHIPS = 1000;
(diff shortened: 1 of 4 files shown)


src/MembershipsSDKAdapter.js
```diff
@@ -89,7 +89,7 @@ function getMembers(sdkMembers) {
   members = sortMeetingMembers(members);
 
   return members.map((member) => ({
-    ID: member.id,
+    ID: member.participant && member.participant.person && member.participant.person.id,
```
cipak, line 92: as we need the `person` 2 times now, let's put it in a var:
```js
const person = (member.participant && member.participant.person) || {};
return {
  ID: person.id,
  orgID: person.orgId,
  ...
};
```

on the other hand, this means introducing the "return" statement which makes the syntax a bit more cluttered. This is one place where the `?.` operator would help:

```js
ID: member.participant?.person?.id,
orgID: member.participant?.person?.orgId,
...
```

@lalli-flores If it's very unlikely that an SDK `member` wouldn't have a defined `participant` property, we can drop the multiple tests and just go with `ID: member.participant.person.id`.

cipak (approved):

src/MembershipsSDKAdapter.js
lalli-flores: I'm not 100% sure but they be not be the same thing, for instance, if a user joins using a Webex device, then it might be a user but not a person (as it is a device). I would need to confirm my assumption.

This is still not mergeable, I left a comment yesterday that hasn't been addressed, but we need to make sure we return a Hydra ID and not a UUID

src/MembershipsSDKAdapter.js
```diff
@@ -89,8 +90,14 @@ function getMembers(sdkMembers) {
...
-    ID: member.id,
-    orgID: member.participant && member.participant.person && member.participant.person.orgId,
+    ID:
+      member.participant
+      && member.participant.person
+      && constructHydraId(hydraTypes.PEOPLE, member.participant.person.id),
+    orgID:
+      member.participant
+      && member.participant.person
+      && constructHydraId(hydraTypes.PEOPLE, member.participant.person.orgId),
```
lalli-flores, line 100:
```suggestion
      && constructHydraId(hydraTypes.ORGANIZATION, member.participant.person.orgId),
```

lalli-flores (approved):
If you can fix the hydra type for orgs it should be good to merge

PR #124, opened 2021-05-17
fix(MeetingSdkAdapter): share button doesn't update the display

alinasuciu:
Fix the local share property to be updated  depending on the sdk event when the local share is stopped.
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-233449

### DIFF
--- a/src/MeetingsSDKAdapter.js
+++ b/src/MeetingsSDKAdapter.js
@@ -2,7 +2,6 @@ import {MeetingsAdapter, MeetingControlState, MeetingState} from '@webex/compone
 import {deconstructHydraId} from '@webex/common';
 import {
   concat,
-  defer,
   from,
   fromEvent,
   merge,
@@ -688,27 +687,24 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
     }
 
     const enableSharingStream = async () => {
-      const [, localShare] = await sdkMeeting.getMediaStreams({sendShare: true});
-
-      this.meetings[ID].localShare = localShare;
-
-      // Due to SDK limitation around local media updates,
-      // we need to emit a custom event for video mute updates
+      // Disable the button while the SDK is enabling sharing
       sdkMeeting.emit(EVENT_MEDIA_LOCAL_UPDATE, {
         control: SHARE_CONTROL,
         state: MeetingControlState.DISABLED,
       });
 
+      const [, localShare] = await sdkMeeting.getMediaStreams({sendShare: true});
+
+      this.meetings[ID].localShare = localShare;
+
       await sdkMeeting.updateShare({stream: localShare, sendShare: true, receiveShare: true});
     };
 
     const disableSharingStream = async () => {
-      // TODO (SPARK-148910): emit should be at the end of the function.
-      // This is a workaround for a bug of async updateShare() call.
-      // The async call may need ~30s to finish execution, or even cannot finish execution.
+      // Disable the control while the SDK is stopping sharing, which could take 30 seconds or more
       sdkMeeting.emit(EVENT_MEDIA_LOCAL_UPDATE, {
         control: SHARE_CONTROL,
-        state: MeetingControlState.INACTIVE,
+        state: MeetingControlState.DISABLED,
       });
 
       await sdkMeeting.updateShare({
@@ -716,8 +712,7 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
         receiveShare: true,
       });
 
-      this.stopStream(this.meetings[ID].localShare);
-      this.meetings[ID].localShare = null;
+      // The rest of the cleanup is done in the handling of the EVENT_LOCAL_SHARE_STOP event emitted by sdkMeeting.updateShare
     };
 
     const resetSharingStream = (error) => {
@@ -815,21 +810,27 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
       }),
     );
 
-    const meetingWithMediaStoppedSharingLocalEvent$ = fromEvent(sdkMeeting, EVENT_LOCAL_SHARE_STOP)
-      .pipe(
-        // TODO (SPARK-152004): This is a workaround for a bug of stopping share when onended event is triggered.
-        flatMap(() => defer(() => this.meetingControls[SHARE_CONTROL].action(ID))),
-        // eslint-disable-next-line no-console
-        tap(() => console.log('EVENT_LOCAL_SHARE_STOP is triggered', this)),
-        map(() => inactiveShare),
-      );
+    const meetingWithMediaStoppedSharingLocalEvent$ = fromEvent(
+      sdkMeeting,
+      EVENT_LOCAL_SHARE_STOP,
+    ).pipe(
+      // eslint-disable-next-line no-console
+      tap(() => console.info('EVENT_LOCAL_SHARE_STOP was triggered', this)),
+      tap(() => {
+        this.stopStream(this.meetings[ID].localShare);
+        this.meetings[ID].localShare = null;
+      }),
+      map(() => inactiveShare),
+    );
 
-    const meetingWithMediaStartedSharingLocalEvent$ = fromEvent(sdkMeeting, EVENT_LOCAL_SHARE_START)
-      .pipe(
-        // eslint-disable-next-line no-console
-        tap(() => console.log('EVENT_LOCAL_SHARE_START is triggered')),
-        map(() => activeShare),
-      );
+    const meetingWithMediaStartedSharingLocalEvent$ = fromEvent(
+      sdkMeeting,
+      EVENT_LOCAL_SHARE_START,
+    ).pipe(
+      // eslint-disable-next-line no-console
+      tap(() => console.info('EVENT_LOCAL_SHARE_START was triggered', this)),
+      map(() => activeShare),
+    );
 
     const sharingEvents$ = merge(
       localMediaUpdateEvent$,
--- a/src/MeetingsSDKAdapter.test.js
+++ b/src/MeetingsSDKAdapter.test.js
@@ -2,19 +2,21 @@ import * as rxjs from 'rxjs';
 import {flatMap, take} from 'rxjs/operators';
 
 import MeetingSDKAdapter from './MeetingsSDKAdapter';
-import createMockSDK, {mockSDKMeeting} from './mockSdk';
+import createMockSDK from './mockSdk';
 
 describe('Meetings SDK Adapter', () => {
   let meeting;
   let meetingID;
   let meetingSDKAdapter;
   let mockSDK;
+  let mockSDKMeeting;
   let target;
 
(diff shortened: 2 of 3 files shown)


src/MeetingsSDKAdapter.js
```diff
@@ -679,7 +679,7 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
     }
 
     const enableSharingStream = async () => {
-      const [, localShare] = await sdkMeeting.getMediaStreams({sendShare: true});
+      const [localShare] = await sdkMeeting.getMediaStreams({sendShare: true});
```
lalli-flores, line 682: doesn't `getMediaStreams` return an array with 2 media streams? one for video and one for screen share, if you get the first one, wouldn't that be the wrong one?

src/MeetingsSDKAdapter.js
```diff
@@ -697,6 +697,11 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
       // TODO (SPARK-148910): emit should be at the end of the function.
       // This is a workaround for a bug of async updateShare() call.
       // The async call may need ~30s to finish execution, or even cannot finish execution.
+      // The stopStream method should be called after updateShare async call
+      // The localShare propery of meeting should be updated after updateShare async call
+      this.stopStream(this.meetings[ID].localShare);
```
lalli-flores, line 702: I'm not sure this is fixing the issue or could potentially cause more problems

src/MeetingsSDKAdapter.js
```diff
@@ -41,8 +40,8 @@ import {
 const EVENT_MEDIA_READY = 'media:ready';
 const EVENT_MEDIA_STOPPED = 'media:stopped';
 const EVENT_STATE_CHANGE = 'meeting:stateChange';
-const EVENT_LOCAL_SHARE_STOP = 'meeting:stoppedSharingLocal';
-const EVENT_LOCAL_SHARE_START = 'meeting:startedSharingLocal';
+const EVENT_LOCAL_SHARE_STOPPED = 'meeting:stoppedSharingLocal';
+const EVENT_LOCAL_SHARE_STARTED = 'meeting:startedSharingLocal';
 const EVENT_REMOTE_SHARE_START = 'meeting:startedSharingRemote';
```
cipak, line 45: seems that these event names also need to be updated. Maybe we can create a different task named "Fix SDK meeting event names used in adapter" just for that

src/MeetingsSDKAdapter.js
```diff
@@ -679,36 +678,36 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
     }
 
     const enableSharingStream = async () => {
-      const [, localShare] = await sdkMeeting.getMediaStreams({sendShare: true});
-
-      this.meetings[ID].localShare = localShare;
-
       // Due to SDK limitation around local media updates,
```
cipak, line 681: maybe fix the comment: disable the button while the SDK is enabling sharing

src/MeetingsSDKAdapter.js
```diff
@@ -679,36 +678,36 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
 
+      const [, localShare] = await sdkMeeting.getMediaStreams({sendShare: true});
+
+      this.meetings[ID].localShare = localShare;
+
       await sdkMeeting.updateShare({stream: localShare, sendShare: true, receiveShare: true});
     };
 
     const disableSharingStream = async () => {
       // TODO (SPARK-148910): emit should be at the end of the function.
```
cipak, line 696: "emit should be at the end..." - not anymore, you can remove this.

you can leave the comment about the call taking 30s, maybe something like:

// disable the control while the SDK is stopping sharing, which make take 30 seconds or more

src/MeetingsSDKAdapter.js
```diff
@@ -679,36 +678,36 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
       });
 
       await sdkMeeting.updateShare({
         sendShare: false,
         receiveShare: true,
       });
 
-      this.stopStream(this.meetings[ID].localShare);
-      this.meetings[ID].localShare = null;
+      // The rest of the cleanup is done in the handling of the STOP event emitted by sdkMeeting.updateShare
```
cipak, line 710: STOP => EVENT_LOCAL_SHARE_STOPPED

src/MeetingsSDKAdapter.js
```diff
@@ -806,21 +805,27 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
-        // eslint-disable-next-line no-console
-        tap(() => console.log('EVENT_LOCAL_SHARE_STOP is triggered', this)),
-        map(() => inactiveShare),
-      );
+    const meetingWithMediaStoppedSharingLocalEvent$ = fromEvent(
+      sdkMeeting,
+      EVENT_LOCAL_SHARE_STOPPED,
+    ).pipe(
+      // eslint-disable-next-line no-console
+      tap(() => console.log('EVENT_LOCAL_SHARE_STOPPED is triggered', this)),
```
cipak, line 813: maybe "was" instead of "is", as we are handling the event here, not triggering it

src/MeetingsSDKAdapter.js
```diff
@@ -806,21 +805,27 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
+      tap(() => {
+        this.meetings[ID].localShare = null;
+        this.stopStream(this.meetings[ID].localShare);
```
cipak, line 816: won't work, the stream was just set to `null` in the line above.

src/MeetingsSDKAdapter.test.js
```diff
@@ -639,6 +639,12 @@ describe('Meetings SDK Adapter', () => {
   });
 
   describe('shareControl()', () => {
+    beforeEach(() => {
+      meetingSDKAdapter.meetings[meetingID] = {
+        ...meeting,
+        localShare: {},
```
cipak, line 645: why is this needed?

src/MeetingsSDKAdapter.test.js
```diff
@@ -717,7 +723,6 @@ describe('Meetings SDK Adapter', () => {
       await meetingSDKAdapter.handleLocalShare(meetingID);
 
       expect(mockSDKMeeting.updateShare).toHaveBeenCalled();
-      expect(meetingSDKAdapter.meetings[meetingID].localShare).toBeNull();
```
cipak, line 720: in order to leave this `expect` in place, we need to wait a little more. You can use https://www.npmjs.com/package/wait-for-expect

alinasuciu: I don't think this should be here anymore since here we call the handleLocalShare method and we check if the disableSharingStream is triggered and this doesn't set the localShare property to null anymore.

src/MeetingsSDKAdapter.test.js
alinasuciu: otherwise it doesn't know about the localShare property

src/MeetingsSDKAdapter.js
alinasuciu: Honestly I didn't know which one in the right one since the naming doesn't correspond with the one from api documentation. I reverted the changes.

src/mockSdk.js
```diff
@@ -14,7 +14,7 @@ export const mockSDKPerson = {
   orgId: 'orgID',
 };
 
-export const mockSDKMeeting = {
+export const createMockSDKMeeting = () => ({
```
cipak, line 17: jsdoc. And better use the "function createMockSDKMeeting" syntax instead of arrow.

cipak (approved):

src/MeetingsSDKAdapter.js
```diff
@@ -679,36 +678,33 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
+
       await sdkMeeting.updateShare({stream: localShare, sendShare: true, receiveShare: true});
     };
 
     const disableSharingStream = async () => {
-      // TODO (SPARK-148910): emit should be at the end of the function.
-      // This is a workaround for a bug of async updateShare() call.
-      // The async call may need ~30s to finish execution, or even cannot finish execution.
+      // Disable the control while the SDK is stopping sharing, which could take 30 seconds or more
+
```
lalli-flores, line 696: You should remove this new line

src/MeetingsSDKAdapter.js
```diff
@@ -806,21 +802,27 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
-        // eslint-disable-next-line no-console
-        tap(() => console.log('EVENT_LOCAL_SHARE_STOP is triggered', this)),
-        map(() => inactiveShare),
-      );
+    const meetingWithMediaStoppedSharingLocalEvent$ = fromEvent(
+      sdkMeeting,
+      EVENT_LOCAL_SHARE_STOPPED,
+    ).pipe(
+      // eslint-disable-next-line no-console
+      tap(() => console.log('EVENT_LOCAL_SHARE_STOPPED was triggered', this)),
```
lalli-flores, line 810: This should probably be informational
```suggestion
      tap(() => console.info('EVENT_LOCAL_SHARE_STOPPED was triggered', this)),
```

src/MeetingsSDKAdapter.js
```diff
@@ -806,21 +802,27 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
-        // eslint-disable-next-line no-console
-        tap(() => console.log('EVENT_LOCAL_SHARE_START is triggered')),
-        map(() => activeShare),
-      );
+    const meetingWithMediaStartedSharingLocalEvent$ = fromEvent(
+      sdkMeeting,
+      EVENT_LOCAL_SHARE_STARTED,
+    ).pipe(
+      // eslint-disable-next-line no-console
+      tap(() => console.log('EVENT_LOCAL_SHARE_STARTED was triggered', this)),
```
lalli-flores, line 823:
```suggestion
      tap(() => console.info('EVENT_LOCAL_SHARE_STARTED was triggered', this)),
```

lalli-flores (commented):
Minor comments but I think this implementation should work 👍🏼
Tests are not passing, however. Could you fix them?

src/mockSdk.js
```diff
@@ -14,90 +14,97 @@ export const mockSDKPerson = {
...
-          participant: {
-            person: {
-              ...mockSDKPerson,
-              id: 'id',
+/**
+ * Creates a mock instance of the Webex SDK Meeting used in unit testing
+ *
+ * @returns {object} mockSDKMeeting Instance
+ */
+export function createMockSDKMeeting() {
```
cipak, line 22: in order to prevent all the indentation changes, can you please switch to the arrow function syntax?

```js
export const createMockSDKMeeting = () => ({
  id: 'meetingID',
  ...,
});

PR #134, opened 2021-06-18
fix(MeetingSdkAdapter): share button doesn't update the display

alinasuciu:
Fix the share banner display to hide when the share screen is closed.

### DIFF
--- a/src/MeetingsSDKAdapter.js
+++ b/src/MeetingsSDKAdapter.js
@@ -859,10 +859,6 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
     ).pipe(
       // eslint-disable-next-line no-console
       tap(() => console.info('EVENT_LOCAL_SHARE_STOP was triggered', this)),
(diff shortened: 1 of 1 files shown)


src/MeetingsSDKAdapter.js
```diff
@@ -750,7 +750,7 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
         receiveShare: true,
       });
 
-      // The rest of the cleanup is done in the handling of the EVENT_LOCAL_SHARE_STOP event emitted by sdkMeeting.updateShare
```
cipak, line 753: let's add a listener (before the `await` above) for the `EVENT_LOCAL_SHARE_STOP` event,

```js
sdkMeeting.once(EVENT_LOCAL_SHARE_STOP, () => {
  this.stopStream(this.meetings[ID].localShare);
  this.meetings[ID].localShare = null;
});

cipak (approved):

PR #137, opened 2021-06-25
feat(MeetingsSdkAdapter): add speaker selection control

lungancatalin:
Add speaker selection control 

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-243984

### DIFF
--- a/scripts/puppeteer.server.js
+++ b/scripts/puppeteer.server.js
@@ -64,6 +64,14 @@ function handleMicrophoneSelect() {
     });
 }
 
+function handleSpeakerSelect() {
+  const switchSpeakerSelect = document.getElementById('switch-speaker');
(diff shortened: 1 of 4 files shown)


scripts/puppeteer.server.js
```diff
@@ -196,6 +205,20 @@ document.getElementById('switch-microphone').addEventListener('change', async ()
   }
 });
 
+document.getElementById('switch-speaker').addEventListener('change', async () => {
+  try {
+    const switchSpeakerSelect = document.getElementById('switch-speaker');
+    const speakerID = switchSpeakerSelect.value;
+    const nodeAudio = document.getElementById('remote-audio');
+    if (nodeAudio?.setSinkId) {
```
cipak, line 213: please move the `if` after the `await`, so if the action fails, we do not change the sink

cipak (approved):

PR #149, opened 2021-07-12
feat(package): configure winston logger

patricia0817:


package.json
```diff
@@ -33,7 +33,9 @@
   },
   "dependencies": {
     "@webex/common": "^1.105.0",
-    "@webex/component-adapter-interfaces": "^1.17.0"
+    "@webex/component-adapter-interfaces": "^1.17.0",
+    "webex": "^1.125.5",
```
cipak, line 37: webex is already a peer dependency, no need to add it here

src/logger.js
```diff
@@ -0,0 +1,27 @@
...
+};
+
+const consoleTransport = new transports.Console(options);
+let fileTransport;
+
+const transportOptions = [consoleTransport];
+
+if (typeof window === 'undefined') {
+  fileTransport = new transports.File({
+    filename: 'logger.log',
```
cipak, line 18: better use an environment variable set in .env: `WEBEX_LOG_FILE`. If and only if set, add the `File` transport

src/logger.js
```diff
@@ -0,0 +1,27 @@
...
+
+const options = {
+  format: format.combine(
+    format.colorize(),
+    format.simple(),
+  ),
+  silent: process.env.NODE_ENV === 'production',
+};
+
+const consoleTransport = new transports.Console(options);
```
cipak, line 11: I would go with:

```js
if (process.env.NODE_ENV !== 'production') {
  activeTransports.push(new transports.Console(options);
}
```

src/logger.js
```diff
@@ -0,0 +1,20 @@
...
+  format: format.combine(
+    format.colorize(),
+    format.simple(),
+  ),
+};
+
+const activeTransports = [];
+
+if (process.env.NODE_ENV !== 'production') {
+  activeTransports.push(new transports.Console(options));
```
cipak, line 13: let's keep the console options near the console instance:
```js
activeTransports.push(new transports.Console({
  format: ...
}));

cipak (approved):

cipak (approved):

PR #153, opened 2021-07-15
feat(MeetingSdkAdapter): remove media before leave meeting

alinasuciu:
Remove emitting of stop media events from leaveMeeting and call the removeMedia method from inside it. 

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-250320

### DIFF
--- a/src/MeetingsSDKAdapter.js
+++ b/src/MeetingsSDKAdapter.js
@@ -572,12 +572,9 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
     try {
       const sdkMeeting = this.fetchMeeting(ID);
 
-      await sdkMeeting.leave();
+      this.removeMedia(ID);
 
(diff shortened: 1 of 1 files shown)


src/MeetingsSDKAdapter.js
```diff
@@ -1362,10 +1358,6 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
         map((event) => this.attachMedia(ID, event)),
       );
 
-      const meetingWithMediaStoppedEvent$ = fromEvent(sdkMeeting, EVENT_MEDIA_STOPPED).pipe(
```
cipak, line 1365: being an SDK event, it might get triggered by the sdk, let's leave this handler here.

What we would need to do would be to see which stream has been stopped, and only set to null that one, not all media streams (audio, video, share). But we can deal with this in a different task.

cipak (approved):

PR #168, opened 2021-07-23
feat(MeetingsSdkAdapter): implement proceed without microphone control

alinasuciu:
Create "proceed without microphone" control display and action. 
Update getLocalMedia method to use chainWith function for chaining observables.

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-250019

### DIFF
--- a/src/MeetingsSDKAdapter.js
+++ b/src/MeetingsSDKAdapter.js
@@ -14,7 +14,6 @@ import {
   concatMap,
   distinctUntilChanged,
   filter,
-  last,
   map,
   publishReplay,
   refCount,
@@ -71,6 +70,7 @@ const SWITCH_CAMERA_CONTROL = 'switch-camera';
 const SWITCH_MICROPHONE_CONTROL = 'switch-microphone';
 const SWITCH_SPEAKER_CONTROL = 'switch-speaker';
 const PROCEED_WITHOUT_CAMERA_CONTROL = 'proceed-without-camera';
+const PROCEED_WITHOUT_MICROPHONE_CONTROL = 'proceed-without-microphone';
 
 // Media stream types
 const MEDIA_TYPE_LOCAL = 'local';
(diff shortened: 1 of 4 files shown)


src/MeetingsSDKAdapter.js
```diff
@@ -165,6 +165,11 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
       action: this.ignoreVideoAccessPrompt.bind(this),
       display: this.proceedWithoutCameraControl.bind(this),
     };
+    this.meetingControls[PROCEED_WITHOUT_MICROPHONE_CONTROL] = {
```
cipak, line 168: blank line before pls (see above)

cipak (approved):

cipak (approved):

src/MeetingsSDKAdapter.js
```diff
@@ -205,10 +213,10 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
         })),
       )
       : of({permission: null, stream: null});
-    const video$ = mediaSettings.sendVideo
+
+    return mediaSettings.sendVideo
```
cipak, line 217: if `sendVideo` is false, the audio messages are not emitted even if `sendAudio` is true

The mapping needs to be done on the `{permission: null, stream: null}` messages as well:

```js
const getStreamWithPermission = (condition, ID, mediaDirection, audioVideo) =>
  condition
    ? this.getStream(ID, mediaDirection, audioVideo)
    :  of({permission: null, stream: null});

const {sendAudio, sendVideo} = mediaSettings;

return getStreamWithPermission(sendAudio, ID, {sendAudio: true}).pipe(
  map(...),
  chainWith((audio) => getStreamWithPermission(sendVideo, ID, {sendVideo: true}).pipe(
    map(...),
  ),
);

File: src/MeetingsSDKAdapter.js [hunk 3]
```diff
@@ -288,6 +289,28 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
...
+   *
+   * @private
+   * @param {boolean} condition Meeting mediaSettings propery condition for sending streams
+   * @param {string} ID Meeting ID
+   * @param {object} mediaDirection A configurable options object for joining a meetings
+   * @param {object} audioVideo audio/video object to set audioinput and videoinput devices
+   * @returns {Observable} Observable that emits local media streams and their user permission status
+   */
+  getStreamWithPermission(condition, ID, mediaDirection, audioVideo) {
+    let mediaStream;
```
cipak, line 303: as it is an observable, you can suffix the '$'
```js
let stream$;
```

cipak (approved):

PR #170, opened 2021-07-26
fix(MeetingsSdkAdapter): refactor tests to use global MediaStream mock object

bogdanmaier28:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-251321

### DIFF
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -1,7 +1,60 @@
 // Mock Web Media APIs
 
-global.MediaStream = jest.fn();
-global.MediaStream.prototype = {
-  getTracks: jest.fn(() => []),
-  removeTrack: jest.fn(),
-};
+global.MediaStream = jest.fn((tracksOrStream = []) => {
+  let tracks;
+
+  if (tracksOrStream instanceof global.MediaStream) {
+    tracks = [...tracksOrStream.getTracks()];
+  } else {
+    tracks = tracksOrStream;
+  }
+
+  for (const track of tracks) {
+    if (!('enabled' in track)) {
+      track.enabled = true;
+    }
+
+    if (!('stop' in track)) {
+      Object.defineProperty(track, 'stop', {
+        value: jest.fn(),
+        enumerable: false,
+      });
+    }
+  }
+
+  return {
+    getTracks: jest.fn(() => tracks),
+    getAudioTracks: jest.fn(() => tracks.filter((track) => track.kind === 'audio')),
+    getVideoTracks: jest.fn(() => tracks.filter((track) => track.kind === 'video')),
+    removeTrack: jest.fn((toRemove) => tracks.filter((track) => track !== toRemove)),
+  };
+});
+
+expect.extend({
+  /**
+   * Custom jest matcher that checks that two media streams have the same tracks
+   *
+   * @param {MediaStream} received
+   * @param {MediaStream} expected
+   * @returns {object}
+   */
+  toMatchMediaStream(received, expected) {
+    const options = {
+      comment: 'Object.is equality',
+      isNot: this.isNot,
+      promise: this.promise,
+    };
+
+    const pass = this.equals(received.getTracks(), expected.getTracks());
+
+    const message = () => `${this.utils.matcherHint('toMatchMediaStream', undefined, undefined, options)}\n
+Expected: ${this.utils.printExpected(expected && expected.getTracks())}
+Received: ${this.utils.printReceived(received && received.getTracks())}`;
+
+    return {
+      actual: received,
+      message,
+      pass,
+    };
+  },
+});
--- a/src/MeetingsSDKAdapter.test.js
+++ b/src/MeetingsSDKAdapter.test.js
@@ -6,7 +6,7 @@ import {
   first,
 } from 'rxjs/operators';
 
-import {mockSDKMediaStreams} from './mockSdk';
+import {createMockSDKMediaStreams} from './mockSdk';
 import {meetingID, createTestMeetingsSDKAdapter} from './MeetingsSDKAdapter/testHelper';
 
 describe('Meetings SDK Adapter', () => {
@@ -14,6 +14,7 @@ describe('Meetings SDK Adapter', () => {
   let meetingsSDKAdapter;
   let mockSDK;
   let mockSDKMeeting;
+  let mockSDKMediaStreams;
   let target;
 
   beforeEach(() => {
@@ -23,6 +24,7 @@ describe('Meetings SDK Adapter', () => {
     mockSDK = meetingsSDKAdapter.datasource;
     mockSDKMeeting = mockSDK.meetings.getMeetingByType('id', meetingID);
     rxjs.fromEvent = jest.fn(() => rxjs.of({}));
+    mockSDKMediaStreams = createMockSDKMediaStreams();
     target = 'target';
   });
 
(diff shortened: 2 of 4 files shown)


jest.setup.js
```diff
@@ -1,7 +1,24 @@
...
+    if (!('enabled' in track)) {
+      track.enabled = true;
+    }
+  }
+
+  return {
+    getTracks: jest.fn(() => tracks),
+    getAudioTracks: jest.fn(() => tracks.filter((track) => track.kind === 'audio')),
+    getVideoTracks: jest.fn(() => tracks.filter((track) => track.kind === 'video')),
+    removeTrack: jest.fn(),
```
cipak, line 22: you can make this work by using an in-place filtering to exclude the received track from the tracks array https://stackoverflow.com/a/49587869/210550

src/MeetingsSDKAdapter.test.js
```diff
@@ -62,22 +64,19 @@ describe('Meetings SDK Adapter', () => {
...
-          localAudio: {
-            stream: mockSDKMediaStreams.localAudio,
-            permission: 'ALLOWED',
-          },
-          localVideo: {
-            stream: mockSDKMediaStreams.localVideo,
-            permission: 'ALLOWED',
-          },
-        });
+        expect(dataDisplay.localAudio.stream.getTracks()).toMatchObject(
```
cipak, line 69: you can define a custom matcher for media streams:
```js
expect(displayData.localAudio.stream).toMatchMediaStream(mockSDKMediaStreams.localAudio);
```
see https://jestjs.io/docs/expect#custom-matchers-api

src/mockSdk.js
```diff
@@ -14,11 +14,19 @@ export const mockSDKPerson = {
...
+  const localVideo = new MediaStream([{kind: 'video'}]);
+
+  return {
+    localAudio,
+    localVideo,
+    localAudioVideo: new MediaStream([...localAudio.getTracks(), ...localVideo.getTracks()]),
+  };
 };
 
+const mockSDKMediaStreams = createMockSDKMediaStreams();
```
cipak, line 28: I guess this is no longer needed

src/MeetingsSDKAdapter.test.js
```diff
@@ -340,18 +332,15 @@ describe('Meetings SDK Adapter', () => {
...
-    let mockMediaStreamInstance;
 
     beforeEach(() => {
-      mockMediaStreamInstance = {
-        getTracks: () => [{stop: mockStop}],
-      };
-
-      global.MediaStream = jest.fn(() => mockMediaStreamInstance);
+      for (const track of mockSDKMediaStreams.localAudio.getTracks()) {
+        track.stop = mockStop;
```
cipak, line 338: I think you can still define the `stop()` function in the mock MediaStream object and avoid matching it by making it not enumerable:

```js
track.defineProperty(
   value: jest.fn(),
   enumerable: false,
);

File: src/MeetingsSDKAdapter.test.js [hunk 5]
```diff
@@ -340,18 +332,15 @@ describe('Meetings SDK Adapter', () => {
...
+      }
     });
 
     test('calls stop() of the stream track', () => {
```
cipak, line 342: > calls stop() on all stream tracks

src/MeetingsSDKAdapter.test.js
```diff
@@ -340,18 +332,15 @@ describe('Meetings SDK Adapter', () => {
...
-      meetingSDKAdapter.stopStream(new MediaStream());
+      meetingSDKAdapter.stopStream(mockSDKMediaStreams.localAudio);
 
       expect(mockStop).toHaveBeenCalled();
```
cipak, line 345:
```js
for(const track of mockSDKMediaStreams.localAudio) {
  expect(track.stop).toHaveBeenCalled();
}

File: jest.setup.js [hunk 7]
```diff
@@ -1,7 +1,55 @@
...
+  };
+});
+
+// Custom Matchers
+
+expect.extend({
+  toMatchMediaStream(received, expected) {
```
cipak, line 36: jsdoc

```js
/**
 * Custom jest matcher that checks that two media streams have the same tracks

File: src/MeetingsSDKAdapter.test.js [hunk 8]
```diff
@@ -179,47 +173,37 @@ describe('Meetings SDK Adapter', () => {
       meetingSDKAdapter.getStream(meetingID, {sendAudio: true})
         .pipe(last())
         .subscribe((localMedia) => {
-          expect(localMedia).toMatchObject({
-            stream: mockSDKMediaStreams.localAudio,
-            permission: 'ALLOWED',
-          });
+          expect(localMedia.stream).toMatchMediaStream(mockSDKMediaStreams.localAudio);
+
```
cipak, line 177: you can lose this empty line, there is a single `expect` of each type in this case

src/MeetingsSDKAdapter.test.js
```diff
@@ -179,47 +173,37 @@ describe('Meetings SDK Adapter', () => {
...
+        stream: mockSDKMediaStreams.localAudioVideo,
       };
-
-      global.MediaStream = jest.fn(() => mockMediaStreamInstance);
     });
 
     test('keeps `localAudio.stream` empty, if the event type is `local` and the audio stream is empty', () => {
       meetingSDKAdapter.meetings[meetingID] = {
-        disabledLocalAudio: {},
+        disabledLocalAudio: mockSDKMediaStreams.localAudio,
```
cipak, line 206: this is a (very) incomplete meeting object, it would be more robust to have a complete object here:
```js
meetingSDKAdapter.meetings[meetingID].disabledLocalAudio = ...;
mmeetingSDKAdapter.meetings[meetingID].localAudio = {stream: null, permission: 'ALLOWED'};
```

cipak: or you can use the same approach in line 352 https://github.com/webex/sdk-component-adapter/pull/170/files#diff-94a8fe68ea1a2cb38f44e0058884a91f48b7e685cbd6d6b1c88152c5d4024627R542

src/mockSdk.js
cipak: yes, because each new mock sdk instance needs to have fresh audio & video media streams. (we will set flags on tracks to false and we don't want this to spill from one test to another)

jest.setup.js
```diff
@@ -1,7 +1,62 @@
...
+      });
+    }
+  }
+
+  return {
+    getTracks: jest.fn(() => tracks),
+    getAudioTracks: jest.fn(() => tracks.filter((track) => track.kind === 'audio')),
+    getVideoTracks: jest.fn(() => tracks.filter((track) => track.kind === 'video')),
+    removeTrack: jest.fn((toRemove) => {
+      tracks.splice(0, tracks.length, ...tracks.filter((track) => !Object.is(track, toRemove)));
```
cipak, line 30: I think it's fine to use the new array without splice. Also, you can use `track !== toRemove`

jest.setup.js
```diff
@@ -1,7 +1,60 @@
...
+    const options = {
+      comment: 'Object.is equality',
+      isNot: this.isNot,
+      promise: this.promise,
+    };
+
+    const pass = this.equals(received.getTracks(), expected.getTracks());
+
+    const message = () => `${this.utils.matcherHint('toMatchMediaStream', undefined, undefined, options)}\n\n
+      Expected: ${this.utils.printExpected(expected)}\n
```
cipak, line 51: for a more helpful output, please try to use:
```js
printExpected(expected?.getTracks())
```

cipak (approved):

PR #200, opened 2021-09-23
fix(MeetingsSdkAdapter): add set timeout of 3 seconds on addMedia

alinasuciu:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-265323


lalli-flores:
We will not move forward with this workaround for the bug.

PR #206, opened 2021-09-29
fix(MeetingsAdapter): add edge case tests

jonniespratley:
- Add conditional check for meeting without `localShare.stream` which covers edge case when trying to leave a meeting and `localShare` is null, it throws an error.

```
Unable to leave from the meeting "SOME-ID" TypeError: Cannot read properties of null (reading 'stream')
    at Pm.removeMedia (webexSDKComponentAdapter.esm.js:15735)
    at Pm.leaveMeeting (webexSDKComponentAdapter.esm.js:15925)
    at wm.action (webexSDKComponentAdapter.esm.js:14860)
    at webex-components.umd.min.js:14
    at o.r.handleClick (index.js:49)
    at onClick (index.js:179)
```

### DIFF
--- a/src/MeetingsSDKAdapter.js
+++ b/src/MeetingsSDKAdapter.js
@@ -393,13 +393,19 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
    * @param {string} ID ID of the meeting to update
    */
   removeMedia(ID) {
-    if (this.meetings && this.meetings[ID]) {
-      this.stopStream(this.meetings[ID].localAudio.stream);
-      this.stopStream(this.meetings[ID].localVideo.stream);
-      this.stopStream(this.meetings[ID].localShare.stream);
-      this.stopStream(this.meetings[ID].disabledLocalAudio);
-      this.stopStream(this.meetings[ID].disabledLocalVideo);
+    const meeting = this.meetings[ID];
+
+    if (!meeting) {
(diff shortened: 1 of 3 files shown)


lalli-flores (commented):
Can you confirm on which scenario would the `localShare.stream` would not exists? From what I can see on the adapter code, [`localShare.stream` is always set/exists](https://github.com/webex/sdk-component-adapter/blob/5bf7f5ff42ebd1d4f185cc0bc27ca454d608c8c8/src/MeetingsSDKAdapter.js#L489)

jonniespratley:
Exact location on where the shape its shifted is https://github.com/webex/react-widgets/blob/master/packages/node_modules/@webex/redux-module-media/src/actions.js#L201

The `eslint-disable` flag is what might have caused the re-assignment to not be flagged.

lalli-flores:
@jonniespratley I think this is on the Space widget then, see the [initial object shape in the adapter](https://github.com/webex/sdk-component-adapter/blob/5bf7f5ff42ebd1d4f185cc0bc27ca454d608c8c8/src/MeetingsSDKAdapter.js#L479) vs what is on the [Space widget](https://github.com/webex/react-widgets/blob/master/packages/node_modules/@webex/redux-module-media/src/actions.js#L211). Please update the Space widget shape to match what is on the adapter.

PR #275, opened 2022-01-17
feat(ActivitiesSdkAdapter): create a method to fetch an activity by id

karinasigartau0798:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-293308

### DIFF
--- a/src/ActivitiesSDKAdapter.js
+++ b/src/ActivitiesSDKAdapter.js
@@ -1,4 +1,8 @@
 import {ActivitiesAdapter} from '@webex/component-adapter-interfaces';
+import {ReplaySubject, defer} from 'rxjs';
+import {map} from 'rxjs/operators';
+
+import logger from './logger';
 
 /**
  * An activity a person performs in Webex.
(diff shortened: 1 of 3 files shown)


src/ActivitiesSDKAdapter.js
```diff
@@ -18,5 +22,58 @@ import {ActivitiesAdapter} from '@webex/component-adapter-interfaces';
...
+      this.activityObservables[ID] = new ReplaySubject(1);
+
+      defer(() => this.fetchActivity(ID)).pipe(
+        map((sdkMessage) => ({
+          ID: sdkMessage.id,
+          roomID: sdkMessage.roomId,
+          text: sdkMessage.text,
+          personID: sdkMessage.personId,
+          created: sdkMessage.created,
+          data: sdkMessage.attachments[0] && sdkMessage.attachments[0].content,
```
cipak, line 63: leave this as `attachments` instead of `data`

and add another field, `card`:

```js
card: attachments && atachments[0] && attachments[0].contentType === 'xxx' && attachments[0].content

File: src/ActivitiesSDKAdapter.js [hunk 2]
```diff
@@ -18,5 +22,58 @@ import {ActivitiesAdapter} from '@webex/component-adapter-interfaces';
...
+   */
+  getActivity(ID) {
+    logger.debug('ACTIVITY', ID, 'getActivity()', ['called with', {ID}]);
+
+    if (!(ID in this.activityObservables)) {
+      // use ReplaySubject cause we don't need to set an initial value
+      this.activityObservables[ID] = new ReplaySubject(1);
+
+      defer(() => this.fetchActivity(ID)).pipe(
+        map((sdkMessage) => ({
```
cipak, line 57: you can destructure here for easier manipulation of these properties:
```js
map(({id, roomId, text, personId, created, attachments}) => ({
```

src/ActivitiesSDKAdapter.js
```diff
@@ -18,5 +22,66 @@ import {ActivitiesAdapter} from '@webex/component-adapter-interfaces';
...
+          personId,
+          attachments,
+          created,
+        }) => ({
+          ID: id,
+          roomID: roomId,
+          text,
+          personID: personId,
+          attachments,
+          card: attachments && attachments[0] && attachments[0].contentType === 'application/vnd.microsoft.card.adaptive' && attachments[0].content,
```
cipak, line 70: this is correct, but there are 3 possible return values for `card` using this expression:
- the card itself
- undefined (if no attachments or no first attachment)
- false (if there is an attachment which is not a card)

it would be slightly better to use a ternary operator, and return either the card or undefined.

src/ActivitiesSDKAdapter.js
```diff
@@ -18,5 +22,66 @@ import {ActivitiesAdapter} from '@webex/component-adapter-interfaces';
...
+          created,
+        })),
+      ).subscribe(
+        (activity) => {
+          logger.debug('ACTIVITY', ID, 'getActivity()', ['emitting activity object', activity]);
+          this.activityObservables[ID].next(activity);
+        },
+        (error) => {
+          logger.error('ACTIVITY', ID, 'getActivity()', 'Error fetching activity', error);
+          this.activityObservables[ID].error(new Error(`Could't find activity with ID "${ID}"`));
```
cipak, line 80: "Couldn't" (missing n), or, better, the more formal "Could not"

src/ActivitiesSDKAdapter.js
```diff
@@ -18,5 +22,66 @@ import {ActivitiesAdapter} from '@webex/component-adapter-interfaces';
 export default class ActivitiesSDKAdapter extends ActivitiesAdapter {
   constructor(datasource) {
     super(datasource);
+
+    this.activityObservables = {};
+  }
+
+  /**
+   * Returns a promise to activity data from Webex.
```
cipak, line 30: Loads activity data from Webex and returns a promise that resolves to an Activity object

src/ActivitiesSDKAdapter.js
```diff
@@ -18,5 +22,66 @@ import {ActivitiesAdapter} from '@webex/component-adapter-interfaces';
...
+   *
+   * @param {string} activityID ID of the activity for which to fetch data
```
cipak, line 32: double space

cipak (approved):

PR #280, opened 2022-02-01
getRoomActivites

taymoork2:
Implement `getRoomActivities`

[SPARK-98332](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-98332)

PR #281, opened 2022-02-01
feat(rooms): add getPreviousActivities method

jonniespratley:
Added unit tests and logic to handle fetching previous activities. 

Example Usage:

```javascript
const webex = new Webex({
  credentials: process.env.WEBEX_ACCESS_TOKEN,
});

const webexSDKAdapter = new WebexSDKAdapter(webex);

// Populate the cache for lookup for roomId
await webexSDKAdapter.datasource.internal.conversation.list({isFavorite: true})

// Subscribe to getPreviousActivites - This will be populated when hasMoreActivites runs
webexSDKAdapter.roomsAdapter.getPreviousActivities(roomId, 10)
    .subscribe({
      next(val) {
        console.log('Fetching past activities');
        console.table(val);
        webexSDKAdapter.roomsAdapter.hasMoreActivities(roomId);
      },
      complete(val) {
        console.log(webexSDKAdapter.roomsAdapter.roomActivities);
        process.exit(0);
      },
      error(val) {
        console.error(val);
      },
    });
```

### DIFF
--- a/src/RoomsSDKAdapter.js
+++ b/src/RoomsSDKAdapter.js
@@ -1,9 +1,11 @@
 import {
+  Subject,
   concat,
   from,
   fromEvent,
   BehaviorSubject,
   Observable,
+  throwError,
 } from 'rxjs';
 import {
   filter,
@@ -28,6 +30,11 @@ import logger from './logger';
 export const ROOM_UPDATED_EVENT = 'updated';
 export const CONVERSATION_ACTIVITY_EVENT = 'event:conversation.activity';
 
+const sortByPublished = (arr) => arr.sort((a, b) => new Date(b.published) - new Date(a.published));
+
+// TODO: Need to remove this once we figure out why we need to pre-cache conversations
+let FETCHED_CONVERSATIONS = false;
+
 /**
  * The `RoomsSDKAdapter` is an implementation of the `RoomsAdapter` interface.
  * This adapter utilizes the Webex JS SDK to fetch data about a room.
@@ -41,6 +48,10 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
     this.getRoomObservables = {};
     this.getRoomActivitiesCache = {};
     this.listenerCount = 0;
+
+    this.activityLimit = 50;
+    this.activitiesObservableCache = new Map();
+    this.roomActivities = new Map();
   }
 
   /**
@@ -51,12 +62,18 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
    * @returns {Room} Information about the room of the given ID
    */
   async fetchRoom(ID) {
-    const {id, title, type} = await this.datasource.rooms.get(ID);
+    const {
+      id,
+      title,
+      type,
+      lastActivity,
+    } = await this.datasource.rooms.get(ID);
 
     return {
       ID: id,
       title,
       type,
+      lastActivity,
     };
   }
 
@@ -138,6 +155,135 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
     }
 
     return this.getRoomObservables[ID];
+  }
+
+  /**
+   * Returns an array of IDs of the most recent activities in a conversation up to the specified limit.
+   *
+   * @param {string} ID ID for the room
+   * @param {string} earliestActivityDate  Get all child activities before this date
+   * @returns {Promise} Resolves with array of activities
+   * @private
+   */
+  async fetchActivities(ID, earliestActivityDate) {
+    const {activityLimit} = this;
+    const conversationId = deconstructHydraId(ID).id;
+
+    logger.debug('ROOM', ID, 'fetchActivities()', ['called with', {
+      earliestActivityDate,
+      activityLimit,
+    }]);
+
+    if (!FETCHED_CONVERSATIONS) {
+      await this.datasource.internal.conversation.list();
+      FETCHED_CONVERSATIONS = true;
+    }
+
+    return this.datasource.internal.conversation.listActivities({
+      conversationId,
+      limit: activityLimit + 1, // Fetch one extra activity to determine if there are more activities to fetch later
+      lastActivityFirst: true,
+      maxDate: earliestActivityDate === null ? undefined : earliestActivityDate,
+    });
+  }
+
+  /**
+   * Returns `true` if there are more activities to load from the room of the given ID.
+   * Otherwise, it returns `false`.
+   *
+   * @param {string} ID ID of the room for which to verify activities.
+   * @returns {boolean} `true` if room has more activities to load, `false` otherwise
+   */
+  hasMoreActivities(ID) {
+    const pastActivities$Cache = this.activitiesObservableCache.get(ID);
+    const {
+      hasMore = true,
+    } = this.roomActivities.get(ID);
+
+    if (!hasMore) {
+      pastActivities$Cache.complete();
+    } else {
+      this.fetchPastActivities(ID);
+    }
+
+    return hasMore;
+  }
+
+  /**
+   * Fetches past activities and returns array of (id, published) objects. Performs side effects
+   *
+   * @param {string} ID The id of the room
+   * @returns null
+   */
+  fetchPastActivities(ID) {
+    const roomActivity = this.roomActivities.get(ID);
+    const {earliestActivityDate} = roomActivity;
+    const room$ = this.activitiesObservableCache.get(ID);
+
+    logger.debug('ROOM', ID, 'fetchPastActivities()', ['called with', {
+      earliestActivityDate,
+    }]);
+
+    if (!ID) {
+      logger.error('ROOM', ID, 'fetchPastActivities()', ['Must provide room ID']);
+      room$.error(new Error('fetchPastActivities - Must provide room ID'));
+    }
+
+    from(this.fetchActivities(ID, earliestActivityDate))
+      .subscribe((data) => {
+        if (!data) {
+          return room$.complete();
+        }
+        roomActivity.hasMore = data.length >= this.activityLimit + 1;
+        const {published} = data.shift();
+        const activityIds = sortByPublished(data).map((activity) => {
+          const {id} = activity;
+
+          roomActivity.activities.set(id, activity);
+
+          return [id, activity.published];
+        });
+
+        roomActivity.earliestActivityDate = published;
+        roomActivity.activityIds.set(published, activityIds.length);
+
+        this.roomActivities.set(ID, roomActivity);
+
+        return room$.next(activityIds);
+      });
+  }
+
+  /**
+   * Returns an observable that emits an array of the next chunk of previous
+   * activity data of the given roomID. If `hasMoreActivities` returns false,
+   * the observable will complete.
+   * **Previous activity data must be sorted newest-to-oldest.**
+   *
+   * @param {string} ID  ID of the room for which to get activities.
+   * @param {number} activityLimit The maximum number of activities to return
+   * @returns {external:Observable.<Room>} Observable stream that emits activity data
+   */
+  getPreviousActivities(ID, activityLimit = 50) {
+    this.activityLimit = activityLimit;
+    const pastActivities$Cache = this.activitiesObservableCache.get(ID) || new Subject();
+
+    if (!ID) {
+      logger.error('ROOM', ID, 'getPreviousActivities()', ['Must provide room ID']);
+
+      return throwError(new Error('getPreviousActivities - Must provide room ID'));
+    }
+
+    if (!this.roomActivities.has(ID)) {
+      this.roomActivities.set(ID, {
+        earliestActivityDate: null,
+        activities: new Map(),
+        activityIds: new Map(),
+      });
+    }
+
+    this.activitiesObservableCache.set(ID, pastActivities$Cache);
+
+    return pastActivities$Cache;
   }
 
   /**
--- a/src/RoomsSDKAdapter.test.js
+++ b/src/RoomsSDKAdapter.test.js
@@ -1,14 +1,17 @@
 import {isObservable} from 'rxjs';
 
 import RoomsSDKAdapter from './RoomsSDKAdapter';
+import mockActivities from './mockActivities';
 import createMockSDK, {mockSDKActivity, mockSDKRoom} from './mockSdk';
 
 describe('Rooms SDK Adapter', () => {
   let mockSDK;
   let roomsSDKAdapter;
+  const roomId = mockSDKRoom.id;
 
   beforeEach(() => {
     mockSDK = createMockSDK();
+    mockSDK.internal.conversation.list = jest.fn(() => Promise.resolve([]));
     roomsSDKAdapter = new RoomsSDKAdapter(mockSDK);
   });
 
@@ -78,6 +81,64 @@ describe('Rooms SDK Adapter', () => {
           created: mockSDKActivity.published,
         });
         done();
+      });
+    });
+  });
+
+  describe('getPreviousActivities() functionality', () => {
+    const getPreviousMock = jest.fn();
+
+    beforeAll(() => {
+      getPreviousMock
+        .mockReturnValueOnce(mockActivities.slice(2))
+        .mockReturnValueOnce(mockActivities.slice(4))
+        .mockReturnValueOnce(null);
+    });
+
+    test('returns an observable', () => {
+      expect(isObservable(roomsSDKAdapter.getPreviousActivities(roomId)))
+        .toBeTruthy();
+    });
+
+    test('completes when all activities have been emitted', (done) => {
+      let itemsCount = 0;
+
+      mockSDK.internal.conversation.listActivities = getPreviousMock;
+      roomsSDKAdapter = new RoomsSDKAdapter(mockSDK);
+
+      roomsSDKAdapter.getPreviousActivities(roomId, 5).subscribe({
+        next(activities) {
+          itemsCount += activities.length;
+        },
+        complete() {
+          expect(itemsCount).toBe(8);
+          done();
+        },
+      });
+
+      roomsSDKAdapter.hasMoreActivities(roomId); // 5
+      roomsSDKAdapter.hasMoreActivities(roomId); // 3
+      roomsSDKAdapter.hasMoreActivities(roomId); // no more
+    });
+
+    test('throws error if no room id is present', (done) => {
+      roomsSDKAdapter.getPreviousActivities().subscribe({
+        next() {},
+        error(e) {
+          expect(e).toEqual(new Error('getPreviousActivities - Must provide room ID'));
+          done();
+        },
+      });
+    });
+
+    test('sets empty roomActivities if no room exists', () => {
+      expect(roomsSDKAdapter.roomActivities.has('room-1')).toBe(false);
+      roomsSDKAdapter.getPreviousActivities('room-1');
+      expect(roomsSDKAdapter.roomActivities.has('room-1')).toBe(true);
+      expect(roomsSDKAdapter.roomActivities.get('room-1')).toStrictEqual({
+        activities: new Map(),
+        earliestActivityDate: null,
+        activityIds: new Map(),
       });
     });
   });
--- a/src/mockActivities.js
+++ b/src/mockActivities.js
@@ -1,0 +1,34 @@
+export default [
+  {
+    id: 'bc2266b2-d6c3-11eb-aef5-6d77908bfcbb',
+    published: '2021-06-25T21:16:12.827Z',
+  },
+  {
+    id: 'c6d88ee0-d6c3-11eb-8b76-fb213805e947',
+    published: '2021-06-23T21:16:30.798Z',
+  },
+  {
+    id: '5f5ad4a0-d7b6-11eb-a8bc-1badfdac8742',
+    published: '2021-06-24T02:13:04.874Z',
+  },
+  {
+    id: 'd33057d0-d831-11eb-a954-81c4a13d6973',
+    published: '2021-06-25T16:56:47.309Z',
+  },
+  {
+    id: '69128f70-d841-11eb-9511-0d3f153f12e6',
+    published: '2021-06-26T18:48:21.223Z',
+  },
+  {
+    id: '81a8f6a0-d841-11eb-a4e5-af1950fe4a84',
+    published: '2021-06-27T18:49:02.474Z',
+  },
+  {
+    id: 'b94c20a0-d841-11eb-9efc-3f4590d442ae',
+    published: '2021-06-28T18:50:35.818Z',
+  },
+  {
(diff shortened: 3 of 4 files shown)


src/RoomsSDKAdapter.js
```diff
@@ -36,6 +38,10 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
 
     this.getRoomObservables = {};
     this.listenerCount = 0;
+
+    this.limit = 50;
```
arun3528, line 42: does this need to be in its own config !?

jonniespratley: no

src/RoomsSDKAdapter.js
```diff
@@ -134,4 +144,119 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+  /**
+   * Invoke request to listActivities on adapter
+   *
+   * @param {object} Params to send to conversation api method
+   * @returns
+   * @private
+   */
+  fetchActivities({roomId, earliestActivityDate}) {
+    const {limit} = this;
+    const conversationId = deconstructHydraId(roomId).id;
```
jonniespratley, line 157: Only works after you use `internal.conversation.list()` as the following methods did not succeed as well:

```
const roomId =
  'Y2lzY29zcGFyazovL3VzL1JPT00vYmMyMjY2YjAtZDZjMy0xMWViLWFlZjUtNmQ3NzkwOGJmY2Ji';
const hydraRoomId = deconstructHydraId(roomId);

console.log(webex.internal.conversation.getConvoUrl(hydraRoomId));
console.log(webex.internal.conversation.getUrlFromClusterId(hydraRoomId));

```

src/RoomsSDKAdapter.js
```diff
@@ -134,4 +144,119 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+        roomActivity.activities.set(id, activity);
+
+        return id;
+      });
+
+      const [{published}] = data;
+
+      earliestActivityDate = published;
+      roomActivity.earliestActivityDate = published;
+      roomActivity.activityIds[earliestActivityDate] = activityIds.length;
```
jonniespratley, line 229: Here we have a map of the date and the number of items. This is where our hasMoreItem checks if we have the number ==== max.

src/RoomsSDKAdapter.js
```diff
@@ -12,6 +13,8 @@ import {
   tap,
 } from 'rxjs/operators';
 import {RoomsAdapter} from '@webex/component-adapter-interfaces';
+
+import {deconstructHydraId} from '@webex/common/dist/uuid-utils';
```
lalli-flores, line 17:
```suggestion
import {deconstructHydraId} from '@webex/common';
```

src/RoomsSDKAdapter.js
```diff
@@ -36,6 +39,10 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
 
     this.getRoomObservables = {};
     this.listenerCount = 0;
+
+    this.limit = 50;
```
lalli-flores, line 43: I would name this a little more explicitly. Not sure "limit" makes sense all on its own
```suggestion
    this.activityChunkSize = 50;
```

src/RoomsSDKAdapter.js
```diff
@@ -134,4 +144,119 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
   }
+
+  /**
+   * Invoke request to listActivities on adapter
+   *
+   * @param {object} Params to send to conversation api method
+   * @returns
+   * @private
+   */
+  fetchActivities({roomId, earliestActivityDate}) {
```
lalli-flores, line 155: Let's have this explicit arguments instead of an object
```suggestion
  fetchActivities(roomID, earliestActivityDate) {
```

src/RoomsSDKAdapter.js
lalli-flores: You should use the cluster information as well, it will hopefully won't require calling `list()` first.

src/RoomsSDKAdapter.js
```diff
@@ -134,4 +144,119 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+    const {limit} = this;
+    const conversationId = deconstructHydraId(roomId).id;
+
+    logger.debug('ROOM', 'fetchActivities()', ['called with', {conversationId, limit, earliestActivityDate}]);
```
lalli-flores, line 159:
```suggestion
    logger.debug('ROOM', 'fetchActivities()', ['called with', {roomID, limit, earliestActivityDate}]);
```

src/RoomsSDKAdapter.js
```diff
@@ -134,4 +144,119 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+
+    let maxDate = earliestActivityDate;
+
+    if (maxDate === null) {
```
lalli-flores, line 163: Just set this as the default value of the argument instead of doing this

src/RoomsSDKAdapter.js
```diff
@@ -134,4 +144,119 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+   * @returns
+   * @private
+   */
+  fetchActivities({roomId, earliestActivityDate}) {
+    const {limit} = this;
+    const conversationId = deconstructHydraId(roomId).id;
+
+    logger.debug('ROOM', 'fetchActivities()', ['called with', {conversationId, limit, earliestActivityDate}]);
+
+    let maxDate = earliestActivityDate;
```
lalli-flores, line 161: Not sure if there is value in doing this

src/RoomsSDKAdapter.js
```diff
@@ -134,4 +144,119 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+
+    if (maxDate === null) {
+      maxDate = undefined;
+    }
+
+    return this.datasource.internal.conversation.listActivities({
+      conversationId,
+      limit,
+      lastActivityFirst: true,
+      maxDate,
```
lalli-flores, line 171:
```suggestion
      maxDate: earliestActivityDate,
```

src/RoomsSDKAdapter.js
```diff
@@ -134,4 +144,119 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
 
     return this.getRoomObservables[ID];
   }
+
+  /**
+   * Invoke request to listActivities on adapter
+   *
+   * @param {object} Params to send to conversation api method
+   * @returns
```
lalli-flores, line 152: Describe the return value

src/RoomsSDKAdapter.js
```diff
@@ -134,4 +144,119 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+    return this.datasource.internal.conversation.listActivities({
+      conversationId,
+      limit,
+      lastActivityFirst: true,
+      maxDate,
+    });
+  }
+
+  /**
+   * Handles checking if there is more activities to fetch.
```
lalli-flores, line 176: Use the [existing description of the interface](https://github.com/webex/component-adapter-interfaces/blob/master/src/RoomsAdapter.js#L81).
Additionally, describe how determining there are more activities left is calculated

src/RoomsSDKAdapter.js
```diff
@@ -134,4 +144,119 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+   *
+   * @param {string} roomId The room id to check
+   * @returns {boolean} True if there is more activities
+   */
+  hasMoreActivities(roomId) {
```
lalli-flores, line 181:
```suggestion
  hasMoreActivities(ID) {
```

src/RoomsSDKAdapter.js
```diff
@@ -36,6 +39,10 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
 
     this.getRoomObservables = {};
     this.listenerCount = 0;
+
+    this.limit = 50;
+    this.roomObservables = {};
+    this.roomActivities = new Map();
```
lalli-flores, line 45: You have to keep in mind that the adapters are singleton objects, so they will be handling multiple rooms. Minor change but makes it clear that this is info from multiple rooms
```suggestion
    this.roomsActivities = new Map();
```

src/RoomsSDKAdapter.js
lalli-flores, line 181: There is also this style pattern in the repo to put all public methods at the end and all privates before public methods.

src/RoomsSDKAdapter.js
```diff
@@ -134,4 +144,119 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+  }
+
+  /**
+   * Fetches past activities and returns array of (id, published) objects.
+   *
+   * @param {string} roomId The id of the room
+   * @param {number} limit The max number of activities to return
+   * @returns {external:Observable.<Activity>} Observable stream that emits activity data
+   */
+  async fetchPastActivities(roomId) {
```
lalli-flores, line 198: Isn't this a private method?

src/RoomsSDKAdapter.js
```diff
@@ -134,4 +144,119 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+      this.roomActivities.set(roomId, roomActivity);
+
+      return roomObservable.next(activityIds);
+    } catch (err) {
+      return roomObservable.error(err);
+    }
+  }
+
+  /**
+   * Handle subscribing to previous activities fetched.
```
lalli-flores, line 240: Use the [existing description of the interface](https://github.com/webex/component-adapter-interfaces/blob/master/src/RoomsAdapter.js#L67). You may add more details specific to the implementation.

src/RoomsSDKAdapter.js
```diff
@@ -134,4 +144,119 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+   *
+   * @param {string} roomId The id of the room
+   * @param {number} count The maximum number of activities to return
+   * @returns {external:Observable.<Room>} Observable stream that emits activity data
+   */
+  getPreviousActivities(roomId, count = 50) {
```
lalli-flores, line 246: Follow naming patterns
```suggestion
  getPreviousActivities(roomID, count = 50) {
```

src/RoomsSDKAdapter.js
```diff
@@ -134,4 +144,119 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+    } catch (err) {
+      return roomObservable.error(err);
+    }
+  }
+
+  /**
+   * Handle subscribing to previous activities fetched.
+   *
+   * @param {string} roomId The id of the room
+   * @param {number} count The maximum number of activities to return
```
lalli-flores, line 243: Brackets indicate that this parameter is optional.

```suggestion
   * @param {number} [count] The maximum number of activities to return
```

src/RoomsSDKAdapter.js
```diff
@@ -134,4 +144,119 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+   * @returns {external:Observable.<Room>} Observable stream that emits activity data
+   */
+  getPreviousActivities(roomId, count = 50) {
+    this.limit = count;
+    if (!roomId) {
+      throw new Error('Must provide roomId!');
```
lalli-flores, line 249: This should throw an error through the observable. Please also make it clearer on the error message where this error is happening

src/RoomsSDKAdapter.js
lalli-flores, line 198: `fetchPastActivities` is not [part of the interface](https://github.com/webex/component-adapter-interfaces/blob/master/src/RoomsAdapter.js#L67) and therefore cannot be public. Where is this method called by the public methods?

src/RoomsSDKAdapter.js
lalli-flores: I'm not sure I understand the logic for determining that you've got the last item

lalli-flores (commented):
Not sure why you picked `fix` as the commit type for this but the changes in this PR are adding new capabilities to the adapters, therefore it should be of type `feat`.
Also, please provide a description of the changes in this PR

src/RoomsSDKAdapter.js
jonniespratley: I have tried every possible way to not have to invoke `list()` to pre-populate the cache before request. I have even stepped thru the code and it seems that something is incorrect. Can you provide an example of how to actually invoke the method in which you are stating.

src/RoomsSDKAdapter.js
```diff
@@ -12,6 +13,8 @@ import {
   tap,
 } from 'rxjs/operators';
 import {RoomsAdapter} from '@webex/component-adapter-interfaces';
+
```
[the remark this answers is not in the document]
jonniespratley: ESLINT auto added it. Are we being picky?

src/RoomsSDKAdapter.js
jonniespratley: Technically its the `limit` of items returned from the request. If we use `activityChunkSize` then it assumes we fetch `500` and chunk `50` Items in sets when invoked, I thought about this as it could save network requests but again it was not specified in the requirements. This is where developers have opinions on how things are named and fact its private so its kind of irrelevant.

src/RoomsSDKAdapter.js
jonniespratley: You could think it like this. A `room` has activities. In order to get a `room` You must specify its `id` using the `get` method you are getting 1 room which has activities. Naming it `roomsActivities` makes it sounds like you want multiple rooms with activities.

src/RoomsSDKAdapter.js
jonniespratley: Deconstructed object is pretty explicit, also makes the API extendable if we wanted to add more options to the underlying request, using object parameters is a best practice as we do not enforce developers to remember the order of parameters or code changes when we add additional required/optional parameters.

src/RoomsSDKAdapter.js
jonniespratley: Since you wanted the name changed to `earliestActivityDate` It made sending deconstructed variables to the underlying odd.

src/RoomsSDKAdapter.js
jonniespratley: Spicy

jonniespratley (commented):
I can see why code sits in PR pending for so long.

src/RoomsSDKAdapter.js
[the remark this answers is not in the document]
lalli-flores: if it's inconsistent, I'll call it out

src/RoomsSDKAdapter.js
lalli-flores: I disagree with your statement:

> This is where developers have opinions on how things are named and fact its private so its kind of irrelevant.

Your team's opinion is not irrelevant and it's a very arrogant attitude.
You're not writing code for yourself, you're writing code for all the people that will read it in the future, naming is very important.

For this case, the context of the the rooms adapter, what is limit? is this for rooms? activities? dates? You definitely need something more descriptive

src/RoomsSDKAdapter.js
lalli-flores: `getPreviousActivities` should only return an observable, if there are errors, the observable should error

lalli-flores:
> I can see why code sits in PR pending for so long.

Is there something taking too long? This PR was opened yesterday and reviewed yesterday

src/RoomsSDKAdapter.js
```diff
@@ -134,4 +153,138 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+  /**
+   * Returns `true` if there are more activities to load from the room of the given ID.
+   * Otherwise, it returns `false`.
+   *
+   * @param {string} ID ID of the room for which to verify activities.
+   * @returns {boolean} `true` if room has more activities to load, `false` otherwise
+   */
+  hasMoreActivities(ID) {
+    const roomObservable = this.roomObservables[ID];
```
lalli-flores, line 199: The convention is to use `$` at the end of the variable name to indicate observable
```suggestion
    const room$ = this.roomObservables[ID];
```

src/RoomsSDKAdapter.js
```diff
@@ -134,4 +153,138 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+   * activity data of the given roomID. If `hasMoreActivities` returns false,
+   * the observable will complete.
+   * **Previous activity data must be sorted newest-to-oldest.**
+   *
+   * @param {string} ID  ID of the room for which to get activities.
+   * @param {number} limit The maximum number of activities to return
+   * @returns {external:Observable.<Room>} Observable stream that emits activity data
+   */
+  getPreviousActivities(ID, limit = 50) {
+    this.limit = limit;
```
lalli-flores, line 271: I still want to see this be named more specifically
```suggestion
    this.activityLimit = limit;
```

src/RoomsSDKAdapter.js
```diff
@@ -134,4 +153,138 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+    const roomObservable = this.roomObservables[ID] || new Subject();
```
lalli-flores, line 272: I like how @taymoork2 [named it on his PR, adding the word "cache"](https://github.com/webex/sdk-component-adapter/pull/280/files#diff-d50aaa2398d276e5ac207ce552bc89420fcf2320576bbce3660e516aa68d26ebR42) to indicate its purpose more clearly.
There are also other room observables also stored, so I would like to see this named more specifically

```suggestion
    const activitiesObservableCache = this.roomObservables[ID] || new Subject();
```
or
```suggestion
    const pastActivities$Cache = this.roomObservables[ID] || new Subject();
```

src/RoomsSDKAdapter.js
```diff
@@ -134,4 +153,138 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+
+    if (!ID) {
+      return throwError(new Error('getPreviousActivities - Must provide room ID'));
```
lalli-flores, line 275: Don't forget to add an error log before returning

src/RoomsSDKAdapter.js
```diff
@@ -134,4 +153,138 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+    }
+
+    if (!this.roomActivities.has(ID)) {
+      this.roomActivities.set(ID, {
+        earliestActivityDate: null,
+        activities: new Map(),
+        activityIds: {},
```
lalli-flores, line 282: Why is one a map and the other an object?

src/RoomsSDKAdapter.js
```diff
@@ -134,4 +153,138 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+    const {earliestActivityDate} = roomActivity;
+    const roomObservable = this.roomObservables[ID];
+
+    logger.debug('ROOM', 'fetchPastActivities()', ['called with', {
+      ID,
+      earliestActivityDate,
+    }]);
+
+    if (!ID) {
+      roomObservable.error(new Error('getPreviousActivities - Must provide room ID'));
```
lalli-flores, line 230: Don't forget to add an error log

src/RoomsSDKAdapter.js
```diff
@@ -134,4 +153,138 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+    }
+
+    return hasMore;
+  }
+
+  /**
+   * Fetches past activities and returns array of (id, published) objects.
+   *
+   * @param {string} ID The id of the room
+   * @returns {Subscription} Observable stream that emits activity data
```
lalli-flores, line 217: A subscription is not an observable, is there value in returning it?

src/RoomsSDKAdapter.js
```diff
@@ -134,4 +153,138 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+
+    if (!ID) {
+      roomObservable.error(new Error('getPreviousActivities - Must provide room ID'));
+    }
+
+    return from(this.fetchActivities({
+      ID,
+      earliestActivityDate,
+    }))
+      .subscribe((data) => {
```
lalli-flores, line 237: I don't think this needs to be an observable, but it is also not a negative thing that it is. The only thing I would change is that if you're going to go the observable route, keep it an observable instead of making it a subscriptions, they are not the same things and cannot be chained

lalli-flores (commented):
I wanted to leave some comments and questions, have yet to test it. You're missing integration tests as well

src/RoomsSDKAdapter.js
jonniespratley: I can make them the same.

jonniespratley:
Rebased, fixed conflicts, updated names to requested, hopefully all issues resolved.

src/RoomsSDKAdapter.test.js
```diff
@@ -1,83 +1,186 @@
...
+            expect(e)
+              .toEqual(new Error('getPreviousActivities - Must provide room ID'));
+            done();
+          },
+        });
+    });
+
+    test('sets empty roomActivities if no room exists', () => {
+      expect(roomsSDKAdapter.roomActivities.has('room-1'))
+        .toBe(false);
```
lalli-flores, line 125:
```suggestion
        .toBeFalsy();
```

src/RoomsSDKAdapter.test.js
```diff
@@ -1,83 +1,186 @@
...
+      roomsSDKAdapter.getPreviousActivities('room-1');
+      expect(roomsSDKAdapter.roomActivities.has('room-1'))
+        .toBe(true);
```
lalli-flores, line 128:
```suggestion
        .toBeTruthy();
```

src/RoomsSDKAdapter.test.js
```diff
@@ -1,83 +1,186 @@
...
+        .toStrictEqual({
+          activities: new Map(),
+          earliestActivityDate: null,
+          activityIds: new Map(),
         });
-        done();
+    });
+  });
+
+  describe('getPreviousActivities() functionality', () => {
```
lalli-flores, line 138: Let's keep titles consistent. I do see that `getRooms` above also has "functionality" in it, but I think this is a one-off that slipped through, the rest of the code base doesn't add that
```suggestion
  describe('getPreviousActivities()', () => {
```

src/RoomsSDKAdapter.test.js
```diff
@@ -1,83 +1,186 @@
...
+    test('returns an observable', () => {
+      expect(isObservable(roomsSDKAdapter.getPreviousActivities(roomId)))
+        .toBeTruthy();
+    });
+
+    test('fetches activities', (done) => {
```
lalli-flores, line 144: Let's have this be more specific to what this is testing
```suggestion
    test('completes when all activities have been emitted', (done) => {
```

lalli-flores (approved):

PR #291, opened 2022-03-07
Added sdk functions for search people and create room

harsh789rai:
1.Added function to get people list based on search query
2. Added function to create room with room title.
3.Added code to test the above two functions in the scripts.

Test cases for the two function have also been added.


src/PeopleSDKAdapter.js
```diff
@@ -184,4 +185,23 @@ export default class PeopleSDKAdapter extends PeopleAdapter {
...
     return this.getPersonObservables[ID];
   }
+
+  /**
+   * Returns an observable that emits people list
+   *
+   * @param {string} query search query
+   * @returns {external:Observable.<Person>} Observable that emits person list based on search query
+   */
+  getPeopleList(query) {
```
cipak, line 195: don't forget to use the logger:
- at the start of the function (getPeopleList() called)
- when emitting something (observer.next)
- when receiving an error (catch)

src/PeopleSDKAdapter.js
```diff
@@ -184,4 +185,23 @@ export default class PeopleSDKAdapter extends PeopleAdapter {
...
+    return new Observable(async (observer) => {
+      try {
+        const peopleList = await this.datasource.people.list({displayName: query});
+
+        observer.next(peopleList.items);
+        observer.complete();
+      } catch (err) {
+        observer.error(new Error('error in fetching data'));
```
cipak, line 203: the original error `err` should be logged (otherwise it's lost) and the message here should be more specific so that it is clear what operation failed

src/PeopleSDKAdapter.test.js
```diff
@@ -119,9 +121,44 @@ describe('People SDK Adapter', () => {
...
+          firstName: 'Webex',
+          lastName: 'Components',
+          avatar: 'avatar',
+          orgId: 'orgID',
+        }]);
+      });
+      done();
+    });
+
+    test('throws error on fetching data', (done) => {
```
cipak, line 144: 'emits an error when the SDK fails to fetch persons'

PR #293, opened 2022-03-08
feat(PeopleSDKAdapter): add method to get people list

harsh789rai:
1. add methods to get people list from webex sdk based on search query
2. add a demo UI under scripts to test the feature
3. add unit test cases

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# [1.106.0](https://github.com/webex/sdk-component-adapter/compare/v1.105.1...v1.106.0) (2022-03-23)
+
+
+### Features
+
+* **ShareControl:** disable share screen on mobile browsers and Internet Explorer browser ([454db60](https://github.com/webex/sdk-component-adapter/commit/454db60a183663a4aeae07763153596a015f52b1))
+
 ## [1.105.1](https://github.com/webex/sdk-component-adapter/compare/v1.105.0...v1.105.1) (2022-03-18)
 
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@webex/sdk-component-adapter",
-  "version": "1.105.1",
+  "version": "1.106.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@webex/sdk-component-adapter",
-  "version": "1.105.1",
+  "version": "1.106.0",
   "main": "dist/webexSDKComponentAdapter.esm.js",
   "module": "dist/webexSDKComponentAdapter.esm.js",
   "scripts": {


src/PeopleSDKAdapter.js
```diff
@@ -184,4 +185,27 @@ export default class PeopleSDKAdapter extends PeopleAdapter {
 
     return this.getPersonObservables[ID];
   }
+
+  /**
+   * Returns an observable that emits people list
+   *
+   * @param {string} query search query
+   * @returns {external:Observable.<Person>} Observable that emits person list based on search query
```
cipak, line 193: Person[] or Array<Person>, as it emits a list of persons, not a single one

src/PeopleSDKAdapter.js
```diff
@@ -184,4 +185,27 @@ export default class PeopleSDKAdapter extends PeopleAdapter {
...
+
+    return new Observable(async (observer) => {
+      try {
+        const peopleList = await this.datasource.people.list({displayName: query});
+
+        observer.next(peopleList.items);
+        logger.debug('PEOPLE', query, 'getPeopleList()', ['emit persons list', peopleList.items]);
+        observer.complete();
+      } catch (error) {
+        logger.error('PEOPLE', query, 'getPeopleList()', ['emits an error when SDK fails to fetch persons list'], error);
```
cipak, line 206: logger.error('PEOPLE', query, 'getPeopleList()', ['error fetching person list'], error);

scripts/puppeteer.server.js
```diff
@@ -273,3 +273,23 @@ document.getElementById('members').addEventListener('click', (event) => {
...
+document.getElementById('query-legend').addEventListener('click', (event) => {
+  event.preventDefault();
+  const query = document.getElementById('query').value;
+
+  try {
+    if (event.target.id === 'query-submit') {
+      webexSDKAdapter.peopleAdapter
+        .getPeopleList(query)
+        .subscribe((items) => {
+          console.log('items', items);
```
cipak, line 286: no need to console.log if you are putting the data in the UI

cipak, line 286: no need to console.log if you are putting the data in the UI

scripts/puppeteer.server.js
```diff
@@ -273,3 +273,23 @@ document.getElementById('members').addEventListener('click', (event) => {
...
+          document.getElementById('people-list').value = JSON.stringify(items, '', 2);
+        }, (error) => {
+          console.log('error --', error);
```
cipak, line 289:
```js
console.error('Error fetching people:', error);

Comment (cipak) [inline scripts/puppeteer.server.js:289]:
```js
console.error('Error fetching people:', error);

File: scripts/puppeteer.server.js [hunk 5]
```diff
@@ -273,3 +273,23 @@ document.getElementById('members').addEventListener('click', (event) => {
...
+        });
+    }
+  } catch (err) {
+    console.log('error', err);
```
cipak, line 293: 'Error fetching people'

It's better to be specific in the log messages, otherwise it's hard to know what failed by just looking at the logs in the console.

cipak, line 293: 'Error fetching people'

It's better to be specific in the log messages, otherwise it's hard to know what failed by just looking at the logs in the console.

cipak: but why the try-catch? what in this code block is expected to throw errors?

cipak: but why the try-catch? what in this code block is expected to throw errors?

scripts/puppeteer.server.js
```diff
@@ -273,3 +273,23 @@ document.getElementById('members').addEventListener('click', (event) => {
...
+  event.preventDefault();
+  const query = document.getElementById('query').value;
+
+  try {
+    if (event.target.id === 'query-submit') {
+      webexSDKAdapter.peopleAdapter
+        .getPeopleList(query)
+        .subscribe((items) => {
+          console.log('items', items);
+          document.getElementById('people-list').value = JSON.stringify(items, '', 2);
```
cipak, line 287: the second parameter of JSON.stringify is an optional replacer function. The specification says it should be `null` if not needed. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#replacer

cipak, line 287: the second parameter of JSON.stringify is an optional replacer function. The specification says it should be `null` if not needed. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#replacer

src/PeopleSDKAdapter.js
```diff
@@ -184,4 +185,27 @@ export default class PeopleSDKAdapter extends PeopleAdapter {
...
   }
+
+  /**
+   * Returns an observable that emits people list
+   *
+   * @param {string} query search query
+   * @returns {external:Observable.<Person[]>} Observable that emits person list based on search query
+   */
+  getPeopleList(query) {
+    logger.debug('PEOPLE', query, 'getPeopleList()', ['called with', {query}]);
```
cipak, line 196: the second param to the logger functions is the entity id (for example, a person id). As we do not have a person id here, you can use 'undefined' or 'null'.

cipak, line 196: the second param to the logger functions is the entity id (for example, a person id). As we do not have a person id here, you can use 'undefined' or 'null'.

src/PeopleSDKAdapter.js
```diff
@@ -184,4 +185,27 @@ export default class PeopleSDKAdapter extends PeopleAdapter {
...
+
+    return new Observable(async (observer) => {
+      try {
+        const peopleList = await this.datasource.people.list({displayName: query});
+
+        observer.next(peopleList.items);
+        logger.debug('PEOPLE', query, 'getPeopleList()', ['emit persons list', peopleList.items]);
+        observer.complete();
+      } catch (error) {
+        logger.error('PEOPLE', query, 'getPeopleList()', ['error fetching person list'], error);
```
cipak, line 206: the third parameter of logger functions is a message (a string) or an array of strings/objects that gets translated into a string by the formatter.

In case you have a single string, there is no need for square brackets, it's just a message.

cipak, line 206: the third parameter of logger functions is a message (a string) or an array of strings/objects that gets translated into a string by the formatter.

In case you have a single string, there is no need for square brackets, it's just a message.

src/PeopleSDKAdapter.test.js
```diff
@@ -119,9 +121,44 @@ describe('People SDK Adapter', () => {
...
+          firstName: 'Webex',
+          lastName: 'Components',
+          avatar: 'avatar',
+          orgId: 'orgID',
+        }]);
+      });
+      done();
+    });
+
+    test('throw error on fetching person list', (done) => {
```
cipak, line 144: getPeopleList does not throw an error, it returns an observable that emits either an array (people list) or an error.

'emits an error if the SDK fails to fetch the person list'

cipak, line 144: getPeopleList does not throw an error, it returns an observable that emits either an array (people list) or an error.

'emits an error if the SDK fails to fetch the person list'

src/mockSdk.js
```diff
@@ -17,6 +17,18 @@ export const mockSDKPerson = {
   orgId: 'orgID',
 };
 
+export const mockSDKPersonList = {
+  items: [{
+    id: 'personIDCurrentUser',
```
cipak, line 22: `id: 'personID'` should be sufficient, no need for a "current user" concept here

cipak, line 22: `id: 'personID'` should be sufficient, no need for a "current user" concept here

lalli-flores (changes requested):
@harsh789rai don't forget to clean up your commits as well, following our [commit guidelines](https://github.com/webex/sdk-component-adapter/blob/master/CONTRIBUTING.md#git-commit). Clean commits are critical, as they determine the version and are the literal [changelog](https://github.com/webex/sdk-component-adapter/blob/master/CHANGELOG.md)

lalli-flores:
Also, if this method is expected to be used by components, it has to be [part of the interface first](https://github.com/webex/component-adapter-interfaces).

PR #300, opened 2022-03-14
fix(RoomsAdapter): fix return value of getActivitiesInRealTime

jonniespratley:
- Fixed the current return value of `getActivitiesInRealTime` to have correct IDs using `constructHydraId`
- Added exported `fromSDKActivity`

### DIFF
--- a/src/RoomsSDKAdapter.js
+++ b/src/RoomsSDKAdapter.js
@@ -3,9 +3,9 @@ import {
   concat,
   from,
   fromEvent,
-  BehaviorSubject,
   Observable,
   throwError,
+  ReplaySubject,
 } from 'rxjs';
 import {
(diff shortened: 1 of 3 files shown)


src/RoomsSDKAdapter.js
```diff
@@ -304,11 +304,11 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
           logger.debug('ROOM', ID, 'getActivitiesInRealTime()', ['received "event:conversation.activity" event', {sdkActivity}]);
 
           const activity = {
```
jonniespratley, line 306: Need to verify if this method should invoke `fromSDKActivity` in the `ActivitiesAdapter` instead of inlining.

PR #301, opened 2022-03-14
feat(cache): add cache module

jonniespratley:
- Added `cache` module to handle caching `activities` and `rooms`
- Purpose of this is to reduce calls to server for `getActivity` method, if `getPastActivities` method caches the returned activities, then when `useActivity` hook is ran cache should lookup and return `activity`.

### DIFF
--- a/src/cache.js
+++ b/src/cache.js
@@ -1,0 +1,124 @@
+import logger from './logger';
+
+/**
+ * @description A module to handle caching data from Webex SDK.
+ * @see {@link https://www.youtube.com/watch?v=ZrtSOTGNqA8}
+ *
+ * @module CacheMeOutside No-pun intented
+ * @returns {CacheMeOutside}
+ */
+class CacheMeOutside {
+  constructor() {
+    this.store = new Map();
+    logger.debug('CacheMeOutside', null, 'init()', []);
+  }
+
+  /**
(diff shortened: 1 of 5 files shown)


src/cache.js
```diff
@@ -0,0 +1,124 @@
+import logger from './logger';
+
+/**
+ * @description A module to handle caching data from webex sdk.
+ *
+ * @module CacheMeOutside No-pun intented
+ * @returns {object} API methods
+ */
+export function CacheMeOutside() {
```
lalli-flores, line 9: Is there a reason to not use the `class` format like the rest of the modules?

src/cache.js
```diff
@@ -0,0 +1,124 @@
+import logger from './logger';
+
+/**
+ * @description A module to handle caching data from webex sdk.
```
lalli-flores, line 4:
```suggestion
 * @description A module to handle caching data from Webex SDK.
 * @see {@link https://www.youtube.com/watch?v=ZrtSOTGNqA8}
```

src/cache.js
jonniespratley: Could do that

jonniespratley:
I will adjust those issues and lint.

taymoork2 (approved):
LGTM

PR #304, opened 2022-03-16
feat(PeopleSdkAdapter): add search people method

harsh789rai:
add searchPeople method to the People sdk adapter to search people that matches the given query.


src/PeopleSDKAdapter.js
```diff
@@ -184,4 +185,28 @@ export default class PeopleSDKAdapter extends PeopleAdapter {
...
+   * Returns an observable that emits a list of people that match the given query.
+   * An empty array is returned when there are no matches.
+   *
+   * @param {string} query Search query
+   * @returns {external:Observable.<Person[]>} Observable that emits person list based on search query
+   */
+  searchPeople(query) {
+    logger.debug('PEOPLE', undefined, 'searchPeople()', ['called with', {query}]);
+
+    return new Observable(async (observer) => {
```
lalli-flores, line 199: You cannot use an `async` function as an observer, you need to use a different operator to start the observable chain that handles `async` otherwise this might lead to race conditions or other unexpected bugs. For promises that need to run afresh every time, you should use [`defer`](https://rxjs.dev/api/index/function/defer)

See my other comment and [getMe()](https://github.com/webex/sdk-component-adapter/blob/master/src/PeopleSDKAdapter.js#L99) from this same adapter for reference

src/PeopleSDKAdapter.js
```diff
@@ -184,4 +185,28 @@ export default class PeopleSDKAdapter extends PeopleAdapter {
...
+      try {
+        const peopleList = await this.datasource.people.list({displayName: query});
```
lalli-flores, line 201: You cannot use an `async` function as an observer, you need to use a different operator to start the observable chain that handles `async` otherwise this might lead to race conditions or other unexpected bugs. For promises that need to run afresh every time, you should use [`defer`](https://rxjs.dev/api/index/function/defer)
```suggestion
        return defer(this.datasource.people.list({displayName: query}));
```

src/mockSdk.js
```diff
@@ -32,6 +32,18 @@ export const mockSDKPerson = {
   orgId: 'orgID',
 };
 
+export const mockSDKPersonList = {
```
lalli-flores, line 35: [`mockSDKPerson`](https://github.com/webex/sdk-component-adapter/blob/master/src/mockSdk.js#L25) already exists, no need to add another mock object

src/mockSdk.js
```diff
@@ -288,6 +300,7 @@ export default function createMockSDK(api = {}) {
     },
     people: {
       get: jest.fn(() => Promise.resolve(mockSDKPerson)),
+      list: jest.fn(() => Promise.resolve(mockSDKPersonList)),
```
lalli-flores, line 303:
```suggestion
      list: jest.fn(() => Promise.resolve({items: [mockSDKPerson]})),
```

lalli-flores:
@harsh789rai, don't forget to update the `@webex/components-adapter-interface` version that contains the `searchPerson` method interface. Please make this change in a separate commit (e.g. `fix(package): update component adapter interfaces to version ZZZ") in addition to the commit for the actual implementation of `searchPeople`

src/mockSdk.js
harsh789rai: I have added the people.list  mock function in the peopleSDKAdapter.test.js file as adding the people.list function in the mock sdk is resulting in the failure of test case in MeetingsSDKAdapter.test.js file.

scripts/index.html
```diff
@@ -40,6 +40,14 @@ <h1 id="page-title">
             <button id="join-meeting">Join Meeting</button>
           </fieldset>
         </form>
+        <form id="search-people-legend">
```
lalli-flores, line 43: Please do not include tooling changes as part of this PR

lalli-flores, line 43: Please do not include tooling changes as part of this PR

src/PeopleSDKAdapter.js
```diff
@@ -184,4 +184,26 @@ export default class PeopleSDKAdapter extends PeopleAdapter {
...
+   *
+   * @param {string} query Search query
+   * @returns {external:Observable.<Person[]>} Observable that emits person list based on search query
+   */
+  searchPeople(query) {
+    return defer(() => this.datasource.people.list({displayName: query}))
+      .pipe(
+        catchError((error) => {
+          logger.error('PEOPLE', undefined, 'searchPeople()', 'error fetching person list', error);
+          throw error;
```
lalli-flores, line 200: According to the description, this should return an empty array, not throw an error
```suggestion
          return [];
```
```suggestion
          throw error;
```

lalli-flores, line 200: According to the description, this should return an empty array, not throw an error
```suggestion
          return [];
```
```suggestion
          throw error;
```

src/PeopleSDKAdapter.js
```diff
@@ -184,4 +184,26 @@ export default class PeopleSDKAdapter extends PeopleAdapter {
...
+        }),
+        map((peopleList) => {
+          logger.debug('PEOPLE', undefined, 'searchPeople()', ['emit persons list', peopleList.items]);
+
+          return peopleList.items;
```
lalli-flores, line 205: Is the format of this list an array of [People objects](https://github.com/webex/component-adapter-interfaces/blob/master/src/PeopleAdapter.js#L6)?

lalli-flores, line 205: Is the format of this list an array of [People objects](https://github.com/webex/component-adapter-interfaces/blob/master/src/PeopleAdapter.js#L6)?

src/PeopleSDKAdapter.js
harsh789rai: Yes it will return an empty array if it does not finds any suitable query matches but it will come under the map function inside the peopleList only as eg:
{"notFoundIds":null,"items":[]}.

And if the sdk returns any error that will come under the catchError method and we will throw it to the function that is subscribing this observable so that they can handle the error accordingly.

src/PeopleSDKAdapter.js
harsh789rai: Yes it contains the list of all the people matching the search query in an array.

scripts/index.html
harsh789rai: I have removed these changes and will include them in a seperate PR once this PR is merged.

src/PeopleSDKAdapter.js
cipak: @harsh789rai there might a difference between a people object as returned by the SDK and a people object as defined in the adapter.

Check if the objects in `peopleList.items` respect the adapter format described in the link Lalli provided https://github.com/webex/component-adapter-interfaces/blob/master/src/PeopleAdapter.js#L6

If not, you should map from sdk format to adapter format, see for example the `fromSDKActivity` function in the Activities adapter which does the same for activities.

src/PeopleSDKAdapter.test.js
```diff
@@ -119,6 +121,54 @@ describe('People SDK Adapter', () => {
...
+            lastName: 'Components',
+            avatar: 'avatar',
+            orgId: 'orgID',
+          },
+        ],
+      }));
+
+      peopleSDKAdapter.searchPeople(query).subscribe((peopleList) => {
+        expect(peopleList).toEqual([{
+          id: 'personIDCurrentUser',
```
cipak, line 146: here, ID should be uppercase, because this is a Person object as defined here:

https://github.com/webex/component-adapter-interfaces/blob/master/src/PeopleAdapter.js#L6

src/PeopleSDKAdapter.js
harsh789rai: I have created a new private function fromSDKPeople to map sdk data to adapter data

src/PeopleSDKAdapter.test.js
harsh789rai: changes done

kishorepmr (approved):

PR #305, opened 2022-03-18
fix(ActivitiesAdapter): fix personID from sdk activity

jonniespratley:
- Fixed issue where personID was not transformed to correct type.


thescripted (approved):
Looks good to me

arun3528 (approved):

PR #307, opened 2022-03-22
feat(RoomsSdkAdapter): add create room method

kishorepmr:


src/RoomsSDKAdapter.js
```diff
@@ -159,6 +161,25 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+  /**
+   * Returns an observable that emits room data of the recently created room
+   * Observable will complete after one emission.
+   *
+   * @param {Room} room  Information about the room to create.
+   * @returns {external:Observable.<Room>} Observable stream that emits room data.
+   * @memberof RoomsAdapter
+   */
+  createRoom(room) {
+    logger.debug('ROOM', room, 'createRoom()', ['called with', {room}]);
```
cipak, line 173: the second param of the logger functions is normally the ID of the resource. Here, we don't have yet an ID, so you can put `undefined` instead. The content of the `room` parameter is already logged in the 'called with' section.

src/RoomsSDKAdapter.js
```diff
@@ -159,6 +161,25 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+
+    return defer(() => this.datasource.rooms.create(room)).pipe(
+      catchError((err) => {
+        logger.error('ROOM', room, 'createRoom()', `Unable to create a room with "${room.title}"`, err);
```
cipak, line 177: maybe not just the title is relevant in case of an error, and if the room.title is "test", the error will read "Unable to create a room with test", which is not very clear. I would suggest changing to:
```js
logger.error('ROOM', undefined, 'createRoom()', ['Unable to create room', {room}], err);

File: src/RoomsSDKAdapter.js [hunk 3]
```diff
@@ -159,6 +161,25 @@ export default class RoomsSDKAdapter extends RoomsAdapter {
...
+   * Observable will complete after one emission.
+   *
+   * @param {Room} room  Information about the room to create.
+   * @returns {external:Observable.<Room>} Observable stream that emits room data.
+   * @memberof RoomsAdapter
+   */
+  createRoom(room) {
+    logger.debug('ROOM', room, 'createRoom()', ['called with', {room}]);
+
+    return defer(() => this.datasource.rooms.create(room)).pipe(
```
cipak, line 175: because this function returns an observable that emits a Room object, you have to make sure you return a Room object, not an SDK room object.

The mapping is currently done in the function "fetchRoom", but I would suggest factoring this simple mapping in a dedicated function called "fromSDKRoom". See how it's done for "fromSDKActivity", for example.

kishorepmr: Added a new private function "fromSDKRoom"

harsh789rai (approved):
LGTM

PR #309, opened 2022-03-25
test(tooling): resusciate integration tests

thescripted:
Get integration tests to work

[SPARK-312424](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-312424)

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,12 +29,12 @@
   "license": "MIT",
   "peerDependencies": {
     "rxjs": "^6.5.4",
-    "webex": "^1.149.2"
+    "webex": "^1.157.0"
   },
   "dependencies": {
     "@babel/plugin-transform-runtime": "^7.16.0",
-    "@webex/common": "^1.149.2",
-    "@webex/component-adapter-interfaces": "^1.23.0",
+    "@webex/common": "^1.157.0",
+    "@webex/component-adapter-interfaces": "^1.26.0",
     "logform": "^2.2.0"
   },
   "devDependencies": {
@@ -49,7 +49,7 @@
     "@rollup/plugin-node-resolve": "^13.0.6",
     "@semantic-release/changelog": "^6.0.0",
     "@semantic-release/git": "^10.0.0",
-    "@webex/test-users": "^1.149.2",
+    "@webex/test-users": "^1.157.0",
     "dotenv": "^8.2.0",
     "eslint": "^7.2.0",
     "eslint-config-airbnb-base": "^14.2.1",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -81,10 +80,14 @@
...
   },
   "jest": {
+    "setupFiles": [
+      "dotenv/config"
+    ],
     "setupFilesAfterEnv": [
       "<rootDir>/jest.setup.js"
     ],
     "preset": "jest-puppeteer",
+    "testEnvironment": "node",
```
lalli-flores, line 90: What's the reason for making node the default instead of JSDOM?

src/PeopleSDKAdapter.integration.test.js
```diff
@@ -1,10 +1,20 @@
...
+ * @jest-environment jsdom
+ */
 import {from} from 'rxjs';
 import {delayWhen, skip} from 'rxjs/operators';
 
-import {createIntegrationTestUser, removeIntegrationTestUser, getPersonHydraID} from './testHelper';
-
 import WebexSDKAdapter from '.';
 
+import {
```
lalli-flores, line 9: Should probably not change unrelated code

src/RoomsSDKAdapter.integration.test.js
```diff
@@ -1,10 +1,19 @@
...
+ * @jest-environment jsdom
+ */
 import {from} from 'rxjs';
 import {delayWhen, skip} from 'rxjs/operators';
 
-import {createIntegrationTestUser, removeIntegrationTestUser} from './testHelper';
-
 import WebexSDKAdapter from '.';
 
+import {
```
lalli-flores, line 9: Should probably not change unrelated code

src/PeopleSDKAdapter.integration.test.js
```diff
@@ -1,10 +1,20 @@
...
+  createIntegrationTestUser,
+  removeIntegrationTestUser,
+  getPersonHydraID,
+} from './testHelper';
+
+// unmock `@webex/common` to avoid `@webex/common` being mocked
```
lalli-flores, line 15:
```suggestion
// Jest auto-mocks by default, avoid `@webex/common` being mocked
```

package.json
thescripted: It looks like a few browser-based queries fails when the testEnvironment is not set to jsdom. Things like `navigator.mediaDevices` appears undefined when running the unit tests. I'm not entirely sure why this is the case.

By default, the testEnvironment is set to node so this just makes the current test behavior explicit. It does make sense for jsdom to be the default although..

lalli-flores: The reason why I would lean towards making JSDOM the default environment for this repository is because the code for the adapter is meant to run on browsers

lalli-flores (approved):

PR #319, opened 2022-04-20
feat(MembershipsAdapter): add addRoomMember method

harsh789rai:
add addRoomMember method to the membershipSDKAdapter.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
   "dependencies": {
     "@babel/plugin-transform-runtime": "^7.16.0",
     "@webex/common": "^1.157.0",
-    "@webex/component-adapter-interfaces": "^1.26.0",
+    "@webex/component-adapter-interfaces": "^1.28.0",
     "logform": "^2.2.0"
   },
   "devDependencies": {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -30,13 +30,13 @@
   "author": "devsupport@webex.com",
   "license": "MIT",
   "peerDependencies": {
-    "rxjs": "^6.5.4",
-    "webex": "^1.157.0"
+    "rxjs": "^6.6.7",
+    "webex": "^1.159.4"
```
kponmuth, line 34: We don't want to update rxjs and webex version Harsh. Is there a need in this PR?

kponmuth (changes requested):

kponmuth:
What about adding membership for a user using his/her email? 

Is that gonna be a separate PR?

package.json
harsh789rai: changes done

harsh789rai:
> What about adding membership for a user using his/her email?
> 
> Is that gonna be a separate PR?

Yes there will be a seperate PR

kponmuth (approved):

PR #334, opened 2024-04-30
fix(meetings): addded password flow for guest joining the meeting

Kesari3008:
In meeting-widgets, the guest was unable to join the meeting due to password flow not being present. Added changes to render the component: WebexMeetingGuestAuthentication which gives the option to enter the password and join the meeting after verifying the password

JIRA: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-505398 

Testcase1: Entered password and joined the meeting as a guest
Console logs: [Widgets_PasswordFlow.log](https://github.com/webex/sdk-component-adapter/files/15169733/Widgets_PasswordFlow.log)

https://github.com/webex/sdk-component-adapter/assets/65543166/0b7a018b-0ae8-4ac7-8951-faa27f7b2395

Testcase2: The guest joined the meeting using the host key.

https://github.com/webex/sdk-component-adapter/assets/65543166/778966dc-7100-4523-b292-c1018e06e9e0

Testcase3: Host joins using host access token

https://github.com/webex/sdk-component-adapter/assets/65543166/3f9c592c-7670-4d95-91e2-62df3e27f946

### DIFF
--- a/src/MeetingsSDKAdapter.test.js
+++ b/src/MeetingsSDKAdapter.test.js
@@ -400,6 +400,56 @@ describe('Meetings SDK Adapter', () => {
       });
     });
 
+    it('returns a new meeting in a proper shape with no passwordStatus', (done) => {
+      meetingsSDKAdapter.fetchMeetingTitle = jest.fn(() => Promise.resolve('my meeting'));
+      meetingsSDKAdapter.getLocalMedia = jest.fn(() => rxjs.of({
+        localAudio: {
+          stream: mockSDKMediaStreams.localAudio,
+          permission: 'ALLOWED',
+        },
+        localVideo: {
+          stream: mockSDKMediaStreams.localVideo,
+          permission: 'ALLOWED',
+        },
+      }));
+
+      mockSDKMeeting.passwordStatus = 'NOT_REQUIRED';
+
+      meetingsSDKAdapter.createMeeting(target).pipe(last()).subscribe((newMeeting) => {
+        expect(newMeeting).toMatchObject({
+          title: 'my meeting',
+          localAudio: {
+            stream: mockSDKMediaStreams.localAudio,
+            permission: 'ALLOWED',
+          },
+          localVideo: {
+            stream: mockSDKMediaStreams.localVideo,
+            permission: 'ALLOWED',
+          },
+          localShare: {
+            stream: null,
+          },
+          passwordRequired: false,
+          remoteAudio: null,
+          remoteVideo: null,
+          remoteShare: null,
+          showRoster: null,
+          settings: {
+            visible: false,
+            preview: {
+              audio: null,
+              video: null,
+            },
+          },
+          state: 'NOT_JOINED',
+          cameraID: null,
+          microphoneID: null,
+          speakerID: '',
+        });
+        done();
+      });
+    });
+
     it('throws error on failed meeting push request', (done) => {
       const wrongTarget = 'wrongTarget';
       const errorMessage = `Unable to create a meeting "${wrongTarget}"`;


Neeraj-swarnkar:
@Kesari3008, I have a few questions:

1. What happens if you enter the wrong password multiple times? How does the system behave?
2. Does the captcha flow get triggered after several failed attempts? [Jira Link]
3. Could you add a recording of the flow where the host tries to join the meeting? I want to make sure the password prompt doesn't appear in that scenario.
4. In the existing test case, could you please verify if a guest can share their screen and whether other participants can view it?

Kesari3008:
> @Kesari3008, I have a few questions:
> 
> 1. What happens if you enter the wrong password multiple times? How does the system behave?
> 2. Does the captcha flow get triggered after several failed attempts? [Jira Link]
> 3. Could you add a recording of the flow where the host tries to join the meeting? I want to make sure the password prompt doesn't appear in that scenario.
> 4. In the existing test case, could you please verify if a guest can share their screen and whether other participants can view it?

@Neeraj-swarnkar  In case the wrong password is entered multiple times then the captcha flow gets triggered. We can see in the console logs that it sets the relevant properties for captch flow but we do not have UI present for the same in components repo. Hence I have created a new Jira to implement the captcha flow which will be picked next once this gets merged.
Sure I will add the recording for host flow for reference.
Sure I can check the share screen functionality and update but it would be something out of the scope of this Jira as this is only for making password flow work so that guest should be able to join the meeting

Kesari3008:
> @Kesari3008, I have a few questions:
> 
> 1. What happens if you enter the wrong password multiple times? How does the system behave?
> 2. Does the captcha flow get triggered after several failed attempts? [Jira Link]
> 3. Could you add a recording of the flow where the host tries to join the meeting? I want to make sure the password prompt doesn't appear in that scenario.
> 4. In the existing test case, could you please verify if a guest can share their screen and whether other participants can view it?

@Neeraj-swarnkar Guest is able to share the screen and other participants can see it

rarajes2:
> Could you add a recording of the flow where the host tries to join the meeting? I want to make sure the password prompt doesn't appear in that scenario.

@Kesari3008, For this case we need to test the flow using host's access token instead of using the host key in the password prompt. 

@Neeraj-swarnkar Please correct me if I misunderstood.

Neeraj-swarnkar:
> > Could you add a recording of the flow where the host tries to join the meeting? I want to make sure the password prompt doesn't appear in that scenario.
> 
> @Kesari3008, For this case we need to test the flow using host's access token instead of using the host key in the password prompt.
> 
> @Neeraj-swarnkar Please correct me if I misunderstood.

@Kesari3008, @rarajes2, 
Yes, the host should join using the host access token. When they attempt to join, they shouldn't encounter the meeting password prompt since they are the host and do not need to enter a password.
@Kesari3008 you can check our KS app once - https://webex.github.io/webex-js-sdk/samples/browser-plugin-meetings/
If host trying to join meeting, verify password is disabled because not needed for host.

Kesari3008:
> > > Could you add a recording of the flow where the host tries to join the meeting? I want to make sure the password prompt doesn't appear in that scenario.
> > 
> > 
> > @Kesari3008, For this case we need to test the flow using host's access token instead of using the host key in the password prompt.
> > @Neeraj-swarnkar Please correct me if I misunderstood.
> 
> @Kesari3008, @rarajes2, Yes, the host should join using the host access token. When they attempt to join, they shouldn't encounter the meeting password prompt since they are the host and do not need to enter a password. @Kesari3008 you can check our KS app once - https://webex.github.io/webex-js-sdk/samples/browser-plugin-meetings/ If host trying to join meeting, verify password is disabled because not needed for host.

@Neeraj-swarnkar @rarajes2 This use-case is an existing flow that is already working. They don't have to enter the password since the WebexMeetingGuestAuthentication component will not render for the host access token case as passwordStatus received from the SDK would not be set to 'REQUIRED'. I will capture a screen recording for this use-case and upload it for reference

Neeraj-swarnkar (approved):
Looks Good.

PR #340, opened 2024-09-27
fix(MeetingsSdkAdapter): handle joinMeeting with empty password

rsarika:
Issue: Not showing password popup when click join meeting after dismissing initial popup.
Fix: added condition to check passwordRequired and empty options to updateMeeting with passwordRequired true

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-557507 

Screenshot 2024-09-30 at 8 31 25 PM
Screenshot 2024-09-30 at 8 31 32 PM

### DIFF
--- a/src/MeetingsSDKAdapter.js
+++ b/src/MeetingsSDKAdapter.js
@@ -635,6 +635,14 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
       const sdkMeeting = this.fetchMeeting(ID);
 
       if (sdkMeeting.passwordStatus === 'REQUIRED') {
+        if (!(options.password || options.hostKey)) {
+          this.updateMeeting(ID, () => (
+            {
+              passwordRequired: true,
+            }));
+
+          return;
+        }
         const res = await sdkMeeting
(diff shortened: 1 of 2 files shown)


Kesari3008:
@rsarika  Please add the screen share of testing for this

src/MeetingsSDKAdapter.js
```diff
@@ -635,6 +635,14 @@ export default class MeetingsSDKAdapter extends MeetingsAdapter {
       const sdkMeeting = this.fetchMeeting(ID);
 
       if (sdkMeeting.passwordStatus === 'REQUIRED') {
+        if (!(options.password || options.hostKey)) {
```
rsarika, line 638: if we dont pass password or hostkey and the meeting requires password then we need not call verify and return from the function saying  passwordRequired: true so that widget shows popup/modal

Shreyas281299: Thanks for the explanation.

Shreyas281299 (approved):

Kesari3008 (approved):

PR #343, opened 2024-10-21
fix(webex): upgrade webex to latest version

Parimala032:
Upgrade webex package version to the latest 2.x as this is causing errors on widgets while resolving webex versions across different packages.

Please do [see this PR](https://github.com/webex/webex-js-sdk/pull/3902#issue-2578353659) for more context that changes made in the file ` MeetingSDKAdapter.js`

- Previously it has been like `name: options:name` , now I am Assigning the `options.name` to the `alias` because the SDK options contain the parameter as the alias, not a name

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "@babel/plugin-transform-runtime": "^7.16.0",
-        "@webex/common": "^2.60.2",
+        "@webex/common": "^2.60.4",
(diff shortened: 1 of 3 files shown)


sreenara (approved):

Shreyas281299 (approved):

mkesavan13 (commented):
Question: Won't this block instant connect? They use the widgets repo and react-widgets repo. Right?

@sreenara @Shreyas281299

mkesavan13 (approved):

PR #344, opened 2024-11-07
fix(MembershipsSdkAdapter):added name property to member mapping in getmembers function

Parimala032:
## This pull request addresses 
This PR enhances the getMembers function by including the `name` property in the returned member objects. By adding the `name: member.name`, each member object now contains the participant's display name in a local sample app of widgets, improving the comprehensiveness of the member data. If the user wanted to pass the `name` according to there wish that will be appear on widgets sample app (or) the system default name will be displayed

please do refer to the screenrecording

https://github.com/user-attachments/assets/5cc7464e-bbc4-470c-a44b-f61e96e93e79

## Summary by CodeRabbit

- **New Features**
	- Enhanced member information by including the member's name in the list returned by the membership function.
  
- **Bug Fixes**
	- Improved error handling for invalid meeting IDs and SDK failures in adding members.

- **Tests**
	- Expanded test coverage for member attributes and refined error handling in the membership-related tests.

### DIFF
--- a/src/MembershipsSDKAdapter.js
+++ b/src/MembershipsSDKAdapter.js
@@ -115,6 +115,7 @@ function getMembers(sdkMembers, meeting) {
     sharing: member.isContentSharing,
     host: !!meeting.meetingInfo.isWebexScheduled && member.isModerator,
     guest: member.isGuest,
+    name:member.name,
   }));
 }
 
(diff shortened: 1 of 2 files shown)


adhmenon (approved):
LGTM. Noticed you have added check in the other repo.

rsarika (approved):

Parimala032:
> LGTM. Noticed you have added check in the other repo.

Yes @adhmenon the check for this is there in the components repo because this function will return the members object that the members object call actually has been there in the components repo so here we just include the name as a property in the members function previously it was not there if we include this property then only the user accepting wished name will be visible on widgets sample app

Parimala032:
@adhmenon  for you question What happens if this value comes as undefined? Do we have an edge check? below is the answer

The current statement has an implicit edge check due to the use of optional chaining and the || operator. It will handle cases where  member.name is undefined by falling back to displayName.

PR #346, opened 2024-12-29
feat(meeting): enable input of password and name while initializing widget

sreenara:
### Summary:
Since we've added the password flow into the Webex Meeting Widget, there is an ask for a developer to pass the meeting password as well as participant name programmatically into the Meeting Widget instead of asking users to enter this once the widget loads up. With this PR, developers will now be able to pass participantName and meetingPasswordOrPin props into the WebexMeetingWidget which will allow users to bypass the authentication and directly land into the Interstitial screen where they can set up audio and video.
Post this, they can directly click "Join Meeting" to start the meeting on the widget.

### Benefits:
This allows developers to programmatically insert the password and name into the widget and avoid having to pass this information to a user in an out-of-band manner.

Check the comments in the code for details on changes.

### Change Summary:
1. All actions now take an object as an argument instead of only meetingID. This has been done for future purposes where we'd like to pass in more properties in the action based on the type of action it is. For instance, for switching camera control, we could pass the new camera device ID.
2. JoinControl takes in meeting pin or password as well as participant name so this can be passed to the adapter while joining the meeting

### Tests:
1. Guest joins meeting from widget without name and password. Prompted for this information before joining the meeting.
2. Guest joins meeting from widget with name and password of the meeting. No prompt and directly lands on interstitial screen, followed by joining the meeting
3. Guest joins meeting from widget with name and host pin of the meeting. No prompt and directly lands on interstitial screen, followed by joining the meeting
4. Registered user joins meeting from widget and is not prompted for password.

Vidcast for test scenarios: https://app.vidcast.io/share/90f8051e-7950-4f79-9d63-670c577f7ea3

### DIFF
--- a/src/MeetingsSDKAdapter/controls/JoinControl.js
+++ b/src/MeetingsSDKAdapter/controls/JoinControl.js
@@ -15,12 +15,15 @@ export default class JoinControl extends MeetingControl {
   /**
    * Calls the adapter joinMeeting method.
    *
-   * @param {string} meetingID  Id of the meeting to join
+   * @param {meetingID, meetingPasswordOrPin}
    */
-  async action(meetingID) {
+  async action({meetingID, meetingPasswordOrPin, participantName}) {
     logger.debug('MEETING', meetingID, 'JoinControl::action()', ['called with', {meetingID}]);
 
-    await this.adapter.joinMeeting(meetingID);
+    await this.adapter.joinMeeting(meetingID, {
+      password: meetingPasswordOrPin,
+      name: participantName,
+    });
   }
 
   /**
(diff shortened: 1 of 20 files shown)


src/MeetingsSDKAdapter/controls/JoinControl.js
```diff
@@ -15,12 +15,15 @@ export default class JoinControl extends MeetingControl {
...
    *
-   * @param {string} meetingID  Id of the meeting to join
+   * @param {meetingID, meetingPasswordOrPin}
    */
-  async action(meetingID) {
+  async action({meetingID, meetingPasswordOrPin, participantName}) {
     logger.debug('MEETING', meetingID, 'JoinControl::action()', ['called with', {meetingID}]);
 
-    await this.adapter.joinMeeting(meetingID);
+    await this.adapter.joinMeeting(meetingID, {
```
sreenara, line 23: This gives widgets the ability to pass any number of arguments into the action argument object

